PR #39, opened 2015-11-28
Phaqui: updated tyggbot/tyggbot.py singleton code

Phaqui:
Just a small update to the singleton class. I don't know if it was working before.. maybe it was.. whatever.


pajlada:
I checked what parts of the code relied on the instance variable, and cleaned that up instead to make the entire thing slightly less spaghetti in this commit: ecc92e169cd8274d8cabae593578280008572ae8

Thanks!

PR #72, opened 2015-12-19
Emote bingo!!

hptk:
So the emote bingo starts, and then the bot automatically picks an emote (out of some legal set) and then the first chatter that guesses the right emote (and only saying that, like spamming all emotes in one message wont work) wins the prize


hptk:
I guess PR is premature, just to get your thoughts and help

hptk:
Alright regarding this there are a few design desisions that need to be made, and I want your input @pajlada. 

* We want a list of standard emotes to chose from, a simple dict in the Tyggbot class would suffice. Does the twitch API suuply a list of global emotes? If not then http://twitchemotes.com/apidocs this site has a list accessible easily. 
* Valid entries in the contest should be only one emote per message, or a message contating a single emote only? I vote for the latter.

PR #100, opened 2015-12-29
Multiaction fallbacks

pajlada:


pajlada:
fucked up some random shit so i accidentally pushed more commits than i should have.
merging anyway, yolo.

PR #102, opened 2015-12-31
Added a VOD link to the Pleblist and !twitterunfollow whisper fix

Nacht123:
__Added a VOD link to the Pleblist History__
**app.py**
Added a template_filter('seconds_to_vodtime')
to convert seconds to VOD Twitch times.

**templates/pleblist_history.html**
Added a VOD link to the pleblist history

It checks the stream_chunk ended time to avoid errors,
if the stream went down and has more than one stream_chunk.

Tested on a small database, need to see this if it works on the big database.

The link can be wrong if the Streamer is pressing pause during the song or
if the streamvod gets the wrong starting time. (maybe some seconds lag)

__Fixed a whisper bug for !twitterunfollow__

### DIFF
--- a/app.py
+++ b/app.py
@@ -21,7 +21,7 @@ from pajbot.models.deck import Deck
 from pajbot.models.command import CommandExample
 from pajbot.models.user import User
 from pajbot.models.duel import UserDuelStats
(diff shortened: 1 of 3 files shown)


pajlada:
Looks good, thank you! :+1:

PR #117, opened 2016-01-06
Lastfm module added.

gigglearrows:


pajlada:
I merged it and made some modifications to fit more with the way I want things to be handled.
The API key and stuff is configurable via the web interface (Which I know you haven't set up yet lol) but I think it works better.
The only thing left would be to make the api key a "password input" so any moderator can't just go in and read the api key.

Thanks for the pull request!

PR #126, opened 2016-01-12
Moved the Bingo system

Nacht123:
Moved the Bingo system to modules
Added a Bingo system for emotes
- __BTTV__ onyl bttv global emotes

- __Twitch__ only bttv global emotes

- __Emotes__ both emotes at the same time

__Added helper__ to the Bingo system
- __Help__
get one random letter
- __Cheat__ 
gets the first letter

This is tested and should work :O

Let me know, if there is something stupid in the code.


pajlada:
Looks good, thanks

PR #131, opened 2016-01-13
Changed some asserts to Pythonic style

ManikDV:
Changed some asserts so all of them now have the same "Pythonic" style.


pajlada:
:+1: Thanks

Issue #132, opened 2016-01-14
fun single game


pajlada:
!roulette
with points
u either win or lose
rigged so the system will always win more Keepo

SnakeTheRipper:
Rigged DansGame

ManikDV:
I think we can close it. @pajlada

pajlada:
@ManikDV Thank you

PR #143, opened 2016-01-20
Roulette minigame + some cosmetic and PEP changes

ManikDV:
#132 
Usage: 
*!rig PERCENT* to set the rig percent (from 0 to 100, when 100 means that user always lose)
*!roulette BET* to play
P.S. Also made a few cosmetic changes.


pajlada:
Thank you sir!

PR #145, opened 2016-01-22
Removed Files

Nacht123:
Updated the Installation Guide for pajbot(Website)

We doesn't need this files,
there will be installed automatically with Semantic.

[New Guide for the Website](/pajlada/pajbot/wiki/Installation-Website)


pajlada:
did i not merge this?

PR #152, opened 2016-01-31
Pyramid settings and pep8 error

Nacht123:
**Updated settings for pyramids**,
we can change the message in the settings now.

Now some people can roleplay with *swiftapples* :D

**Fixed pep8 whitespace for raffles**
Fixed some pep8 whitespace error's,
that i did before :P


pajlada:
looks awesome, thank you!

PR #176, opened 2016-02-19
Update roulette's command example

joaopms:
The old description was from !duel


joaopms:
#173 takes care of this ;-;

pajlada:
@joaopms #173 gives the wrong message

PR #178, opened 2016-02-20
!roulette all-in

T-Brawl:
!allin is shorter to type obviously but this is the least "changing code" solution I guess.

I don't wanna break everything BabyRage


pajlada:
oops sorry accidentally implemented it

T-Brawl:
Never lucky BabyRage

PR #179, opened 2016-02-21
Added description for duel commands

TalVivian:
Just filling a few blank gaps on the website :)


TalVivian:
Tested the change without swapping the file. My bad.
Should be good right now :)

pajlada:
Looks good, thank you! Reply with your twitch username if you want some points in forsens or my chat :sunglasses:

PR #204, opened 2016-02-27
Added a stop-button for playsound on the webpage

dersand:
Added a stop-button for the play sound module. Very trivial stuff, only html and a tiny script inside the onclick attribute.

[How it looks like](http://i.imgur.com/rzl806s.png)

[Using this javascript](https://stackoverflow.com/questions/14834520/html5-audio-stop-function)


dersand:
Also, first ever pull request. Pajlada please don't permban me if i do something wrong BabyRage

pajlada:
Pull request is fine, it's just not something I think I want in that form.
If you could make it so the Play button changes text to "Stop" when the sample is playing, and turns back into a "Play"-button when it's done, that'd be something I'd be interested in.
Thank you!

dersand:
Please try it before you actually commit this. I'm not 100% it works, more like 80%, okay, 90%.
Also travis thing passed, i'm not sure entirely what this means, but i guess it's good.

[Here is human readable code on what i did.](https://i.imgur.com/q6dymbN.png)

PR #215, opened 2016-03-03
eightball.py

nuuls:
Kappa Keepo PogChamp xD 123


dextermb:
Truly wonderful work! :fire:

Nacht123:
**sub emotes** and **user bttv emotes** are in the list

pajbot/modules/eightball.py
```diff
@@ -109,7 +115,7 @@ def eightball_command(self, **options):
         message = options['message']
 
         if message and len(message) > 0:
-            phrase = random.choice(self.phrases)
+            phrase = random.choice(self.phrases) + " " + random.choice(self.emotes)
```
pajlada: Using string concatenation like that is generally frowned upon.
You should instead keep phrase as it was before:
`phrase = random.choice(self.phrases)`
and then do
`emote = random.choice(self.emotes)`
and change the bot.me message to include {phrase} and then {emote} instead

pajlada:
Looks good, thank you!

PR #262, opened 2016-03-25
Advanced play/stop button functionality for Playsounds module

dersand:
This adds functionality for the play sound button on the playsounds module.

[This is my own test case and was shown on stream](https://i.imgur.com/QMj3HBG.png)

### DIFF
--- a/pajbot/modules/tokencommands/playsound.py
+++ b/pajbot/modules/tokencommands/playsound.py
@@ -181,7 +181,7 @@ class PlaySoundTokenCommandModule(BaseModule):
             </script>"""
         local_script = """<script>
             var elem{0.command}=document.getElementById('btnTogglePlay{0.command}');
-            var snd{0.command} = new Audio('{0.href}');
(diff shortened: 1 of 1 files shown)


dersand:
I copied playsounds.py, deleted and reforked the fork. 

 \#Programmer BrokeBack

Nacht123:
Doesn't work with `!#playsound amazing` it should be reuploaded.

This **link** will **break the command page**
(no scrolling for playsound and the button doesn't work with this link.)

dersand:
@Nacht123 Yes, that right. It looks like it's because of pajlada's url [involves a single quote. ](http://i.imgur.com/Vh0u4qz.png)
Is a single quote a valid file name character? If so, `{0.href}` needs to be escaped.

OR

`var snd{0.command} = new Audio('{0.href}');` (line 184)
can be changed to
`var snd{0.command} = new Audio("{0.href}");` (Notice the double quotes inside the paramter)

PR #270, opened 2016-03-29
Fixes for Trivia / Highlights / Roulette Issues

### LINKED ISSUE #168: Boolean option to whisper small roulette results instead of typing them in chat

### LINKED ISSUE #211: Make roulettes whisperable
Should we reply back in whispers? or should we still output the results in chat.
maybe that should be a setting? pajaW

### LINKED ISSUE #235: Trivia - answer's bug
**Remove** html python stuff in answers
```
02:10  Snusbot No one could answer the trivia! The answer was "<i>Gilligan\'s Island</i>"
02:13  Snusbot No one could answer the trivia! The answer was ""Big Bertha""

02:17  Snusbot A new question has begun! In the category "stop, thief!", the question/hint/clue is ""
02:18  Snusbot Here's a hint, "<i>2001: A Space ___________"
02:18  Snusbot: swiftapples got the answer right! The answer was <i>2001: A Space Odyssey</i> FeelsGoodMan

02:18  Snusbot No one could answer the trivia! The answer was "the A\'s"

# EDIT

19:28 TimeoutBanUnban Emsarrev: the Rijksmuseum (in amsterdam)
19:28 TimeoutBanUnban Thedude2k1: Amsterdam
19:28 Snusbot: Emsarrev got the answer right! The answer was the Rijksmuseum (in Amsterdam) FeelsGoodMan
```
<br/>
__EDIT__
Trivia from other people, to see how they did it :santa: 
https://github.com/nyxsys/python-toys/blob/master/trivia/triviaquestions.py
https://github.com/jgrosswm/jasper_personal/blob/master/jasper/client/modules/Trivia.py

### LINKED ISSUE #249: Remove/hide old highlights
Automatically remove highlights from the page if they're older than 2months as they're not stored by twitch ( http://blog.twitch.tv/2014/08/update-changes-to-vods-on-twitch/ ). Second option would be to be able to create regular highlight but I don't think there's an API for that

I think there should maybe just be a warning text next to a highlight that's more than 2 months old, saying the vod link might be broken cuz it's too old.

Nacht123:
### Removes old highlights from the website
Fixes #249

### Trivia: Reture if there is a link in the question
Fixed #235

### New Roulette settings

Fixed #168
Boolean if you allow roulettes in whispers.

Fixed #211
Options for result output
- Chat
- Whisper
- Chat if X points else whisper

And you can now change the message

### DIFF
--- a/pajbot/modules/roulette.py
+++ b/pajbot/modules/roulette.py
@@ -23,6 +23,28 @@ class RouletteModule(BaseModule):
     CATEGORY = 'Game'
     SETTINGS = [
             ModuleSetting(
+                key='message_won',
+                label='Won message | Available arguments: {bet}, {points}, {user}',
+                type='text',
+                required=True,
+                placeholder='{user} won {bet} points in roulette and now has {points} points! FeelsGoodMan',
+                default='{user} won {bet} points in roulette and now has {points} points! FeelsGoodMan',
+                constraints={
+                    'min_str_len': 10,
+                    'max_str_len': 400,
+                }),
+            ModuleSetting(
+                key='message_lost',
+                label='Lost message | Available arguments: {bet}, {points}, {user}',
+                type='text',
+                required=True,
+                placeholder='{user} lost {bet} points in roulette and now has {points} points! FeelsBadMan',
+                default='{user} lost {bet} points in roulette and now has {points} points! FeelsBadMan',
+                constraints={
+                    'min_str_len': 10,
+                    'max_str_len': 400,
+                }),
+            ModuleSetting(
                 key='rigged_percentage',
                 label='Rigged %, lower = more chance of winning. 50 = 50% of winning. 25 = 75% of winning',
                 type='number',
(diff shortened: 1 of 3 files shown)


pajbot/modules/roulette.py
```diff
@@ -158,10 +209,32 @@ def roulette(self, **options):
...
+                else:
+                    bot.whisper(user.username, self.get_phrase('message_won', **arguments))
         else:
-            bot.me('{0} lost {1} points in roulette and now has {2} points! FeelsBadMan'.format(user.username_raw, bet, user.points_available()))
+            if self.settings['options_output'] == '1. Show results in chat':
+                bot.me(self.get_phrase('message_lost', **arguments))
+            if self.settings['options_output'] == '2. Show results in whispers':
+                bot.whisper(user.username, self.get_phrase('message_lost', **arguments))
+            if self.settings['options_output'] == '3. Show results in chat if it\'s over X points else it will be whispered.':
+                if points <= -self.settings['min_show_points']:
```
pajlada: 😱

Nacht123:
Yeah it's ugly but if you use options you need the whole text or? :octocat:

pajlada:
I meant this line:

`if points <= -self.settings['min_show_points']:`

Why not just

`if points  > self.settings['min_show_points']:`

For  the options you're right, it has to look like that for now. It's pretty ugly. Also not good if you ever want to change the message a bit, fix a spelling mistake or something.

pajlada:
Looks great though, I love the simple fixes for the highlight/trivia fixes 😊

Nacht123:
Not sure if this would work without changing the code more in a different way,
since this are only the **negative points** there.

- Points are **always negative**
- self.settings['min_show_points'] are **always positive**

it would **never** get the **if setting**, it would be **always** the **else setting**.


This **would work if** the code would be **structured for every option**,
**but i structured it after win and lose**

pajlada:
Ah yes, you're right of course 😊

PR #359, opened 2019-03-23
Use whisper instead of say in eval_from_file

RAnders00:


pajlada:
Looks good to me, thank you very much!

PR #374, opened 2019-05-28
Open directory containing files

DatGuy1:


static/images/mkfavico.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+cd "${0%/*}"
```
RAnders00: I would appreciate a small couple-word comment above this line explaining what it does

PR #380, opened 2019-06-02
Remove nl_pos and point_pos from dispatch.py (not used anymore)

RAnders00:
The substitutions `$(user:num_lines)`, `$(user:num_lines_rank)`, `$(user:points)`, `$(user:points_rank)` replace these dispatch functions

As far as I could see none of the other code uses these functions, but there could be bots out there still using these instead of the substitution based commands


pajlada:
Looks good to me

PR #383, opened 2019-06-07
Fix timeout event not working

DatGuy1:


pajlada:
I haven't tested it, but considering you're the only user of the bot that I know who uses this feature, I'll just assume you know what you're doing :smiley:

PR #384, opened 2019-06-08
[POC] Docker version of pajbot

KasperHelsted:
I have tried to make an proof-of-concept deployment using docker.
I took some creative freedom and made some changes to some minor files.
Since there were some hard coded URI's, such as the `redis` host and `twitch-api-v3-proxy`, since everything is now running on an internal docker network, they need to communicate over that, like it's multiple machines.

I have chosen to use an unofficial docker image for MariaDB, since this version allows for the creation of multiple databases, while running as a single instance.

Since the implementation of this bot is so seamless and easy to setup i had to create two instances in the `docker-compose.yml` in order to run the `main.py` and `app.py`, OMEGALUL.

I have enabled persistent storage to all containers, so data are save between changes of image, like updating, on the bot the "only" file right now is the config i am saving between updates.
The MariaDB and redis have all their save data stored in persistent volumes.

I have choose to rebuild the MariaDB with a new `my.cnf` file, this was the easiest fix for a the `2006, mysql server has gone away` since this image ships with a timeout time of 600, and the normal default timeout is 28800.

However this implementation seems to have some problems as it is right now...
Since this implementation of this fine bot is so nice and seamless, when starting the first time, the `pajbot_web` container will crash, this is due to the database scheme not existing. So this needs restart,
The `pajbot_bot` will also continuously crash until the token have been created in the redis container.
Also the image on the website, an example could be [logo_pajlada_tn.png](https://paj.pajlada.se/static/images/logo_pajlada_tn.png), since the files in the container should not be manually edited after deployment, maybe hotlink to twitch image?


docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+    depends_on:
+      - mysql
+      - redis
+      - kraken
+    command: ["python3", "main.py", "--config", "configs/config.ini"]
+
+  kraken:
+    image: zwb3/twitch-api-v3-proxy:latest
+    restart: always
+    command: ["/opt/twitch-api-v3-proxy-boot/bin/twitch-api-v3-proxy", "--server.address=0.0.0.0", "--clientId=${CLIENTID}"] # CLIENTID is enviroment variable on server
```
RAnders00: I don't know docker this well, but is it possible that on a server not configured with an iptables firewall that the apiproxy container becomes reachable from the outside?

pajbot/apiwrappers.py
```diff
@@ -248,7 +249,7 @@ def __init__(self, client_id=None, oauth=None, strict=True):
         APIBase.__init__(self, strict)
 
         self.base_url = "https://api.twitch.tv/api/"
-        self.kraken_url = "http://127.0.0.1:7221/kraken/"
+        self.kraken_url = f"http://{os.environ.get('KRAKEN_URL', '127.0.0.1:7221')}/kraken/"
```
RAnders00: We currently want to remain compatible with Python 3.5 (because the documented/recommended deployment is Debian stable), which doesn't support the `f""` format syntax. To remain compatible this needs to be changed to classic style format.

pajbot/managers/redis.py
```diff
@@ -17,7 +18,7 @@ class RedisManager:
 
     @staticmethod
     def init(**options):
-        default_options = {"decode_responses": True}
+        default_options = {"decode_responses": True, "host": os.environ.get("REDIS", "localhost")}
```
RAnders00: There's already a `[redis]` configuration section in the config, is it possible to move the docker-specifica there?

It would just be:

```none
[redis]
host=a.b.c.d
```

pajbot/models/stream.py
```diff
@@ -91,7 +92,7 @@ def fetch_video_url_stage1(self):
             data = self.bot.twitchapi.get(
                 ["channels", self.bot.streamer, "videos"],
                 parameters={"broadcasts": "true"},
-                base="http://127.0.0.1:7221/kraken/",
+                base=f"http://{os.environ.get('KRAKEN_URL', '127.0.0.1:7221')}/kraken/",
```
RAnders00: Same applies here with the format

pajbot/web/routes/base/login.py
```diff
@@ -28,7 +28,7 @@ def init(app):
         consumer_key=app.bot_config["webtwitchapi"]["client_id"],
         consumer_secret=app.bot_config["webtwitchapi"]["client_secret"],
         request_token_params={"scope": "user_read"},
-        base_url="http://127.0.0.1:7221/kraken/",
+        base_url=f"http://{os.environ.get('KRAKEN_URL', '127.0.0.1:7221')}/kraken/",
```
RAnders00: Same applies here with the format

requirements.txt
```diff
@@ -73,6 +73,7 @@ pyflakes==2.1.1
 PyHamcrest==1.9.0
 pylast==3.1.0
 PyMySQL==0.9.3
+mysqlclient==1.4.2.post1
```
RAnders00: Why do we need yet another mysql client?

pajbot/web/utils.py
```diff
@@ -70,7 +71,9 @@ def download_logo(client_id, streamer):
 
     twitchapi = TwitchAPI(client_id)
     try:
-        data = twitchapi.get(["users", streamer], base="http://127.0.0.1:7221/kraken/")
+        data = twitchapi.get(
+            ["users", streamer], base=f"http://{os.environ.get('KRAKEN_URL', '127.0.0.1:7221')}/kraken/"
```
RAnders00: Same applies here with the format

RAnders00: Also in line with the `[redis]` configuration thing, if the redis config thing is possible, then it would be nice to configure this in the config as well (maybe under [twitchapi] and [webtwitchapi]). When the api proxy was introduced this was a quick fix because the (back then) deadline for v3 was coming soon.

RAnders00: Also, a `KRAKEN_URL` should contain the `http://` and `/kraken` bits too, if you just want to ship kraken domain and port then it should be `KRAKEN_HOST` or `APIPROXY_HOST`

docker-compose.yml
KasperHelsted: This is just proof-of-concept, but all the containers are talking to each other over their own internal "network" so unless you expose the ports via the docker-compose, or one of the containers are compromised, they should be safe as far as i know.
Also what i usually do is, lock down almost all incoming ports, and serve all through nginx, via proxy_pass

pajbot/managers/redis.py
KasperHelsted: I could not find  the redis in the `config.example.ini`, and i was not aware how i could easily access the config files from other classes.
Also since the containers are running on an internal network, instead of connecting to a local ip `a.b.c.d`, they should connect to the `name` of the service requested, as seen in these types of environment variables can be seein in the docker-compose file.

requirements.txt
KasperHelsted: It may just be a fluke, but while testing i kept getting errors when communicating with MariaDB and using the `mysql+pymysql`. i tried and google and people were just using the `mysql`, but in order for me to use that i had to install this client.

pajbot/managers/redis.py
KasperHelsted: Also, since the `redis`, and `twitch-api-v3-proxy`, URI's have to be the same for all bot's and web's would it not be smarter to use something like environment variables in the docker-compose file?

RAnders00: Nothing wrong with environment variables, it's just that are not used at all in the bot currently. I might look into this further (maybe we can implement a shared config) because docker deployment is something i support and want to see, and I think the same applies for @pajlada.

docker-compose.yml
RAnders00: That's fair! That's perfect then.

RAnders00:
Also just generally commenting, there has been work done on the bot to move to unix socket files in places where TCP ports are not strictly required:

- The websocket server can listen on unix sockets
- The recommended install procedure makes the bot connect to the mysql server using a unix socket and unix socket (unix user) authentication
- The webserver (uwsgi) can listen on unix sockets
- Technically (although we haven't done this yet) we could even connect to redis over a unix socket.

Is it possible to support any of these with the docker config?

pajbot/managers/redis.py
KasperHelsted: The thing is, those items i have choose to use environment variables for, are settings which should be the same for all the `bot`'s and `web`'s defined, i can see the use for using the config where settings have to differ, like the `mysql` connection, where each of them have to use a different database.
But as far as i understood from spamming pajlada, all the current bots and webs use the same `redis` databse and `twitch-api-v3-proxy` instance?

RAnders00:
Also another thing I didnt see here - the webservice uses a socket to talk to the bot, does that work?

KasperHelsted:
> Also another thing I didnt see here - the webservice uses a socket to talk to the bot, does that work? It's defined in the config as:
> 
> ```ini
> ; This socket lets the bot process and the web process communicate
> ; This is NOT the socket of the uwsgi process to proxy web requests to
> [sock]
> sock_file = /srv/pajbot/kkonatestbroadcaster-bot.sock
> ```

Ahh i had NO idea of what that was, ill fix that tomorrow xD

KasperHelsted:
> Also just generally commenting, there has been work done on the bot to move to unix socket files in places where TCP ports are not strictly required:
> 
> * The websocket server can listen on unix sockets
> * The recommended install procedure makes the bot connect to the mysql server using a unix socket and unix socket (unix user) authentication
> * The webserver (uwsgi) can listen on unix sockets
> * Technically (although we haven't done this yet) we could even connect to redis over a unix socket.
> 
> Is it possible to support any of these with the docker config?
> 
> I'm afraid it's currently not linked anywhere in this repo, but here is a recent documentation write-up and config set, which includes these setups: https://github.com/pajbot/pajbot1-doc
> 
> The reason this is nice is because pajbot1 is often used in parallel to power multiple streamers, and it's nice to not have to maintain a number of ports for each streamer, and instead have them in `/srv/pajbot/%streamer-%service.sock`

Yeah this should be do-able. As far as i understand about sockets, we should be able to do this via volume shares.

KasperHelsted:
> ```ini
> sock_file
> ```

When looking at it, it should be able to use unix socket by using:
`[sock]
sock_file = /opt/www/configs/kkonatestbroadcaster-bot.sock`
Since the `/opt/www/configs` folder is shared between the containers running the bot, so this should in theory work.

pajbot/managers/redis.py
RAnders00: Yea, they all share it.

RAnders00:
Here's another one I'd like to address and request feedback from both @pajlada and @KasperHelsted - I'm a bit dissatisfied with the number of small miscellanious files in the project root of the project folder, and this PR grows this number even more. Is it feasible to move all relevant data (e.g. the `Dockerfiles`, `docker-compose.yml`, mysql configuration, etc.) into a `docker/` subdirectory and then put a small `README.md` inside `docker/` documenting the docker installation process/files?

KasperHelsted:
That should be fixable, ill look into that tomorrow.

pajlada:
After the most recent change, the mysql installation doesn't work

Console output:
```
WARNING: The CLIENTID variable is not set. Defaulting to a blank string.
Building mysql
Step 1/2 : FROM ganjaaa/mariadb:latest
 ---> 2f5c34b814b0
Step 2/2 : COPY "./mysql/my.cnf" "/etc/mysql/my.cnf"
ERROR: Service 'mysql' failed to build: COPY failed: stat /var/lib/docker/tmp/docker-builder983396931/mysql/my.cnf: no such file or directory
```

PR #426, opened 2019-07-24
Docker experiments master merge

RAnders00:
Continuing #384


RAnders00:
I would just like to get the new requirement out and then we can get this merged

PR #431, opened 2019-08-06
Make repspam module configurable, fixes #419 and #428

### LINKED ISSUE #419: Make repspam module configurable
- how many repetitions to allow
  - the timeout length
  - notify whisper?
  - bypass level

### LINKED ISSUE #428: Ignore chatterino character/small spam suffixes in repspam module
Maybe complete #419 first so the module can then be adjusted to work on 4 maximum repetitions again.

RAnders00:
Also cleans out the old suffix tree node stuff and fixes #428 (ignoring chatterino character and dot spam)

### DIFF
--- a/pajbot/modules/repspam.py
+++ b/pajbot/modules/repspam.py
@@ -125,4 +125,4 @@ class RepspamModule(BaseModule):
 
             # found a group of equally repeating words (a repeating spam) that repeats more than allowed
             self.bot.timeout_user_once(source, self.settings["timeout_length"], "No repetitive messages OMGScoods")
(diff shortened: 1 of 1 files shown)


RAnders00:
I believe this is either identical to the old module in terms of behaviour or at least *very* similar. I believe the only difference to be that the old module would only look at the group of words that repeat the most (e.g. the highest repetition word group), this implementation considers lower-repeating groups too (because the highest repetition word group could not feature more than `min_unique_words_in_spam` words, and a lower repeating word group might fulfill that criterion

PR #433, opened 2019-08-06
Added or_else filter, partially addresses #411

### LINKED ISSUE #411: Add substitution to fall back to tb:streamer if nothing is present
`$(user;1:username|orElse($(tb:broadcaster)))` maybe?

RAnders00:
The reason the first commit doesn't close #411 is because a command defined as `$(strictargs:0-1|or_else($(tb:broadcaster)))` actually evaluates to `󠀀$(strictargs:0-1|or_else(randers))` in chat, and is not further expanded.

### DIFF
--- a/docs/variables.txt
+++ b/docs/variables.txt
@@ -21,17 +21,22 @@ $(args:0-1|urlencode) - urlencode the first parameter
 
 Available filters:
         available_filters = {
-                'strftime': _filter_strftime,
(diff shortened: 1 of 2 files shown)


RAnders00:
I decided it's not possible to easily add recursive expansion to filter arguments, so i've just added `or_streamer` and `or_broadcaster` for now.

PR #463, opened 2019-08-27
PostgreSQL migration

RAnders00:


RAnders00:
Chatters microservice will have to be updated. I suggest the following SQL script:

```postgresql
BEGIN;

CREATE TEMPORARY TABLE (username TEXT) ON COMMIT DROP;

INSERT INTO chatters
VALUES ('randers'),
       ('pajlada'),
       ('leppunen');

INSERT INTO "user"(username, username_raw, level, points, subscriber, minutes_in_chat_online,
                   minutes_in_chat_offline)
    (SELECT chatters.username,
            chatters.username,
            100,
            points_base,
            FALSE,
            CASE WHEN stream_online THEN update_interval ELSE 0 END,
            CASE WHEN NOT stream_online THEN update_interval ELSE 0 END
     FROM chatters)
ON CONFLICT (username) DO UPDATE SET
    points = "user".points + CASE WHEN "user".subscriber THEN round(points_base * sub_multiplier) ELSE points_base END,
    minutes_in_chat_online  = "user".minutes_in_chat_online + CASE WHEN stream_online THEN update_interval ELSE 0 END,
    minutes_in_chat_offline = "user".minutes_in_chat_offline + CASE WHEN NOT stream_online THEN update_interval ELSE 0 END;

COMMIT;
```

PR #475, opened 2019-09-06
A few minor fixes

DatGuy1:
These are what I can think of, if I missed any I'll make another PR

### DIFF
--- a/pajbot/modules/roulette.py
+++ b/pajbot/modules/roulette.py
@@ -5,7 +5,6 @@ from numpy import random
 
 import pajbot.exc
 import pajbot.models
-import pajbot.utils
 from pajbot import utils
 from pajbot.managers.db import DBManager
 from pajbot.managers.handler import HandlerManager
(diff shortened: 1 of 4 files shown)


pajlada:
I would rather have the log messages removed instead of commented out

pajbot/modules/roulette.py
```diff
@@ -170,7 +169,7 @@ def roulette(self, **options):
 
         msg_split = message.split(" ")
         try:
-            bet = pajbot.utils.parse_points_amount(user, msg_split[0])
```
RAnders00: this was broken?

pajbot/modules/roulette.py
```diff
@@ -5,7 +5,6 @@
 
 import pajbot.exc
 import pajbot.models
-import pajbot.utils
 from pajbot import utils
```
RAnders00: should have normalized other usages of `utils.` too, since i see the import remained here

PR #477, opened 2019-09-07
Cleanup requirements

Leppunen:


.gitignore
```diff
@@ -27,3 +27,4 @@ userdispatch.py
 venv/
 
 .idea/
+.vscode/
```
RAnders00: LF at the end of the file missing here

install/venvinstall.sh
```diff
@@ -8,14 +8,24 @@ if [ ! -d install ]; then
...
 
 # Upgrade pip
-pip install pip --upgrade
+./venv/bin/pip install pip --upgrade
 
 # Install requirements.txt
-pip install -r requirements.txt
+./venv/bin/pip install -r requirements.txt
+
+if [ $CIRCLECI ]; then
```
RAnders00: should be `$CI` to make it more portable

requirements-dev.txt
```diff
@@ -0,0 +1,21 @@
...
+pluggy==0.12.0
+py==1.8.0
+pycodestyle==2.5.0
+pyflakes==2.1.1
+pyparsing==2.4.2
+pytest==5.1.2
+six==1.12.0
+toml==0.10.0
+wcwidth==0.1.7
+zipp==0.6.0
```
RAnders00: I think `pytest` should be in `requirements-dev.txt`, and CircleCI should use `venvinstall.sh --dev`. Then you wouldn't have to detect `$CI` in `venvinstall.sh`.

Leppunen: Cannot be done until Python 3.5 is removed from CircleCI as it throws an error while trying to install black which is in the dev requirements.

install/venvinstall.sh
Leppunen: This is now fixed.

.gitignore
Leppunen: This is now fixed.

PR #482, opened 2019-09-09
load module before enable

RAnders00:
a bit late but this branch was lingering around the repo


pajlada:
gj pajlada looks good

PR #484, opened 2019-09-10
Create pull_request_template.md, check changelog style in CI

RAnders00:
I did this on my fork to validate it works (that the PR template is picked up by GitHub)


RAnders00:
I renamed `SLOTMACHINE-THINKING.md` to `SLOTMACHINE-THINKING.txt` since it currently is not really markdown-formatted: https://github.com/pajbot/pajbot/blob/edf808f7f2a873db40ed53280fa22b6f0ad0396a/docs/SLOTMACHINE-THINKING.md

.circleci/config.yml
```diff
@@ -35,24 +35,40 @@ jobs:
...
+      - run:
+          name: Install apt repository for Node.js 12
+          command: curl -sL https://deb.nodesource.com/setup_12.x | bash -
+      - run:
+          name: Install Node.js 12
+          command: apt-get install nodejs -y
+      - checkout
+      - run:
+          name: Check markdown style with prettier
+          command: npx prettier@^1.18.2 --check --prose-wrap=always **/*.md
```
pajlada: I'd like the prose-wrap parameter removed, I don't like the line-wrapping things it does.
Also, moving this to a script under `scripts/` would be dank, so you can just run this before committing (or even put it in your pre-commit hook!)

PR #485, opened 2019-09-10
Remove !myuptime command/feature

RAnders00:
I want to leave this open until we merge PostgreSQL into master, then I have my shitty migrations system available, and can create a "migration" that deletes the personal uptime data from redis too.

changelog has been updated


RAnders00:
needs testing as of now

PR #498, opened 2019-09-26
Pajlada add minimum level to configure module

### LINKED ISSUE #108: [Module] Minimum level requirement to configure
Each module should specify a minimum level requirement to configure or enable/disable it.
By default, this level should be 500.

pajlada:
Implement "minimum level to configure" support for modules.

This will do 3 things:
- When a moderator is viewing the list of modules, the configure button will be disabled if they are not high enough level to configure the module.
- If a moderator tries to view the /edit/<moduleid> page, a 403 error will be thrown explaining that they aren't allowed to configure this module.
- If a moderator tries to send a post request to configure the module to the /edit/<moduleid> page, the same 403 error page will be thrown.

Fixes #108

Pull request checklist:

### DIFF
--- a/templates/errors/403.html
+++ b/templates/errors/403.html
@@ -4,4 +4,7 @@
 {% block body %}
 <h1>403</h1>
 <p>You shouldn't be here.</p>
+{% if extra_message %}
+<p>{{ extra_message }}</p>
+{% endif %}
 {% endblock %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Remember to bring your dependencies up to date with
   users that changed their twitch name. (Level 2000 only).
   `./scripts/transfer-{all,sql,redis}` scripts have been removed.
 - Minor: Removed `!reload` command since it did nothing.
+- Minor: Modules can now be configured to only allow users above a certain level to configure them. #108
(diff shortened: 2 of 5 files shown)


templates/errors/403.html
```diff
@@ -4,4 +4,7 @@
 {% block body %}
 <h1>403</h1>
 <p>You shouldn't be here.</p>
+{% if extra_message %}
+<p>{{ extra_message }}</p>
```
RAnders00: maybe rename to `error_detail_message`?

RAnders00:
Maybe the "minimum level to configure" should get per-setting support too? - as an optional keyword argument to the Setting() constructor.

RAnders00:
Also, should it be possible to disable or enable a module if the user does not have `CONFIGURE_LEVEL`?

templates/errors/403.html
pajlada: Agreed

pajlada:
https://github.com/pajbot/pajbot/pull/498#issuecomment-535588609
I don't think CONFIGURE_LEVEL should control enabling or disabling. That should probably be its own level limit setting if anything.

pajlada:
https://github.com/pajbot/pajbot/pull/498#issuecomment-535588405
Agreed. Maybe limit to be able to read setting as well? For secret api keys and what not

PR #501, opened 2019-09-27
Unify user model

### LINKED ISSUE #443: Errors in main thread actions exit the bot
E.g. in functions scheduled with `bot.execute_now`.

RAnders00:
- Removed code for paid timeout discounts (usertags were removed previously so this feature was broken anyways)
- Use f-strings from python 3.6 where possible
- Errors in the main thread (after init) no longer exit the bot, fixes #443
- Remove shitty StreamElements points sync implementation
- Removed $(notify:xxx) substitution, removed test overlay code stuff that was never finished
- bot.ban, bot.timeout, bot.untimeout, bot.unban now take a User instead of login name
- Use / instead of . to prefix commands in IRC since that's the documented approach
- Added cache_bulk_fetch_fn to ResponseCache class
- Twitch Helix API Wrapper: Implement really rudimentary rate limit throttling (only works with one thread)
- Twitch Helix API Wrapper: Make caching system more efficient
- Twitch Helix API Wrapper: Add bulk user data fetch function
- Admin log has been moved into the DB
- Cleaned up tags parsing, tags is now always a dict, and tags values are empty strings where they should be now
- Cleaned up the transaction control in DB migrations a bit
- Added some foreign key constraints to the DB where they were missed
- All in-memory structures storing data about users primary-key by user ID now
- Replaced DuelStatsManager with static methods on DuelStats
- added ON DELETE CASCADE/SET NULL where appropriate to make it possible to delete users from the DB without foreign key conflicts
- User data is no longer stored in redis (data has been moved into SQL from redis)
- Removed UserSQLCache
- Reimplemented points_rank (and num_lines_rank) using materialized views
- Users in the DB are now primary-keyed by their twitch ID (an automatic migration exists)
- Removed namechange module (no longer needed)
- Removed lots of usages of options["bot"], options["source"], etc.. when handling commands
- Reworked duel module to use user IDs and display names
- Reduced/removed some code duplication in followage check module
- Reworked hsbet to primary-key its data structures by user ID
- Removed debug logging spam from link checker
- mass ping module now uses the DB to check for known/existing users
- Warnings are now indexed by user ID, not name
- removed some unreachable code
- replaced some long nested IFs with early-out returns
- Added once=bool parameter to bot.timeout
- Added a login_last_updated parameter to the users table to be able to handle multiple users sharing the same login (username recycling by Twitch)
- Renamed user.username to user.login and user.username_raw to user.name (in line with how the new helix Twitch API does it)
- Added an easy-to-use redis dumping script (for bot data)
- Removed UserManager completely
- Removed in-memory data about users
- Remove email api (?) seems like it was completely unused
- probably other things i've missed
- Added `!reload subscribers` command for level >= 100 (cooldown is 2 minutes)
- Subscribers reload now happens 100% on action queue
- Added option to refresh chatters back into the bot
- Refactored the HSBet module to only have its state inside the DB (Removed persistence in redis, removed user.debts in-memory data)
- User.timed_out and User.timeout_end are properties stored inside the DB now

Pull request checklist:


RAnders00:
todo: finish bulk subscribers update stuff

RAnders00:
Todo: Might be possible to put the chatters update stuff back into the bot. The update can happen asynchronously like in subscribers update

pajlada:
I might be messing with this PR a lot with #503 - take that into consideration when reviewing #503

RAnders00:
@pajlada don't worry I'll merge master into this to get it back into mergeable state should it break

RAnders00:
So I placed the changelog in its own section, so it's easier to then update should the 1.37 release happen (#507).

pajlada:
this is a branch I'd like to have rebased into separate branches, it's currently hard for me go through and check off things that work/don't work

- "Removed code for paid timeout discounts (usertags were removed previously so this feature was broken anyways)" - should be done in a separate PR
- "Use / instead of . to prefix commands in IRC since that's the documented approach" - should probably be done in a separate PR
- "Admin log has been moved into the DB" - should probably? be done in a separate PR
- "Removed lots of usages of options["bot"], options["source"], etc.. when handling commands" - should probably be a separate PR?
- "Reduced/removed some code duplication in followage check module" - should be a separate PR
- "Removed debug logging spam from link checker" - should be a separate PR
- "Renamed user.username to user.login and user.username_raw to user.name (in line with how the new helix Twitch API does it)" - Helix naming is inconsistent, I've seen user_name and display_name, but I'm fine with login and name keys
- "Remove email api (?) seems like it was completely unused" - should be a separate PR
- "Added option to refresh chatters back into the bot" - should be a separate PR

RAnders00:
This pull request is now best viewed one commit at a time.

image

pajbot/modules/basic/checkmod.py
```diff
@@ -16,24 +18,21 @@ class CheckModModule(BaseModule):
...
+                    bot.say(f"{username} was not found in the user database")
         else:
             user = source
 
-        if user:
-            if user.moderator:
-                bot.say("{0} is a moderator PogChamp".format(user.username_raw))
-            else:
-                bot.say("{0} is not a moderator FeelsBadMan (or has not typed in chat)".format(user.username_raw))
+        if user.subscriber:
```
pajlada: should be user.moderator

PR #548, opened 2019-10-12
Add instructions for redis dump to changelog

RAnders00:
Pull request checklist:

image


RAnders00:
I also created a list of redis keys the bot currently uses

https://gist.github.com/RAnders00/b5dd8077f466b50571f83b17e86f9341

I guess some of these could be in the dump script as well?

pajlada:
adding the new redis keys and what-not to the redis dump script should probably be done after the next release

PR #560, opened 2019-10-13
Remove dependency on numpy

RAnders00:
numpy usages were replaced with calls to the standard library

Pull request checklist:

### DIFF
--- a/pajbot/modules/slotmachine.py
+++ b/pajbot/modules/slotmachine.py
@@ -1,7 +1,7 @@
 import logging
 from collections import Counter
 
-from numpy import random
+import random
 
 import pajbot.exc
 import pajbot.models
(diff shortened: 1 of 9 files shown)


RAnders00:
The only notable difference is that numpy's `randint` excludes the `max` parameter (`[min; max)`), while the standard python library counts it as inclusive (`[min; max]`).

pajbot/modules/slotmachine.py
```diff
@@ -24,7 +24,7 @@ def pull_lol(low_tier_emotes, high_tier_emotes, bet, house_edge, ltsw, htsw, ltb
     for e in high_tier_emotes:
         slot_options += [e]
 
-    randomized_emotes = random.choice(slot_options, 3).tolist()
+    randomized_emotes = random.sample(slot_options, 3)
```
pajlada: should be `random.choices(slot_options, k=3)` as random.sample returns a unique list of choices

see https://docs.python.org/3/library/random.html#random.choices and https://docs.python.org/3/library/random.html#random.sample

PR #572, opened 2019-10-20
small migration changes and notes

pajlada:
squash and merge pls

after this is in, we will make a version

Pull request checklist:

### DIFF
--- a/pajbot/migration_revisions/db/0004_unify_user_model.py
+++ b/pajbot/migration_revisions/db/0004_unify_user_model.py
@@ -226,15 +226,18 @@ def up(cursor, bot):
         # create Server-side cursor
         cursor.execute('DECLARE all_users CURSOR FOR SELECT id, login FROM "user" ORDER BY id FOR UPDATE')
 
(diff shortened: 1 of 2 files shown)


pajbot/migration_revisions/db/0004_unify_user_model.py
```diff
@@ -118,16 +226,19 @@ def safe_to_int(input):
         # create Server-side cursor
```
RAnders00: no need for `if users_count > 0` anymore, it was meant so we don't show a progress bar on new databases

pajbot/migration_revisions/db/0004_unify_user_model.py
```diff
@@ -113,36 +221,36 @@ def safe_to_int(input):
     # migrate users to ID
     cursor.execute('SELECT COUNT(*) FROM "user"')
     users_count = cursor.fetchone()[0]
+    offset = 0
```
RAnders00: would be nicer if offset=0 was closer to the loop

RAnders00:
:rage:

PR #585, opened 2019-10-21
Change time_since function to only use plural when the number is equal to 1.

aidenwallis:
Right now anything below 1, such as 0.8 seconds prints as 0.8 second in chat. This doesn't make sense, and propose instead only use the singular when it is exactly equal to 1.

Pull request checklist:


RAnders00:
Thank you! 🎉

PR #596, opened 2019-10-23
Fix pleblist table

pajlada:
fix pleblist table structure

Pull request checklist:


pajbot/migration_revisions/db/0010_make_pleblist_user_id_nullable.py
```diff
@@ -0,0 +1,27 @@
...
+    cursor.execute(
+        """
+    SELECT attnotnull
+    FROM pg_attribute
+    WHERE
+        attrelid = 'pleblist_song'::regclass
+    and attname = 'user_id'
+    """
+    )
+    pleblist_song_user_id_column_unnullable = cursor.fetchone()[0]
```
RAnders00: Please add a comment on why this necessary and why we can't directly "ALTER TABLE" - note that the code for revision 4 was changed after it was released, and so this migration will only run for "older" databases

pajbot/migration_revisions/db/0010_make_pleblist_user_id_nullable.py
```diff
@@ -0,0 +1,27 @@
...
+
+def up(cursor, bot):
+    # This query selects the "nullable" state of the pleblist_song.user_id column
+    cursor.execute(
+        """
+    SELECT attnotnull
+    FROM pg_attribute
+    WHERE
+        attrelid = 'pleblist_song'::regclass
+    and attname = 'user_id'
```
RAnders00: I don't agree with this indentation, I would indent this line as well

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unversioned
 
+- Major: `pleblist_song` table mistakenly wiped any non-user-connected rows in v1.38, this version will make the `pleblist_song` table able to contain non-user-connected rows again.
+
+  Here I will put instructions for how to load your old database into a temporary one and getting those lost rows back:
```
RAnders00: change "temporary one" to "temporary schema"

pajbot/migration_revisions/db/0010_make_pleblist_user_id_nullable.py
pajlada: Added an additional line above the query itself

pajbot/migration_revisions/db/0010_make_pleblist_user_id_nullable.py
RAnders00: Also I would capitalize AND

PR #608, opened 2019-10-25
Prototype new install docs

RAnders00:
I've tried to remove the stuff that is not directly relevant to pajbot.

Pull request checklist:


RAnders00:
The result can be viewed here: https://github.com/pajbot/pajbot/tree/randers00-new-install-docs/install-docs

pajlada:
Looks nice, @alazymeme can use this to set up a bot and let me know if he has any feedback

alazymeme:
First step doesn't work on Ubuntu v19.10 - not sure if you're adding a specific version to the guide though: https://pastebin.com/raw/mcJ1gkCJ

alazymeme:
I'm not sure if I should be using the review function, as conflicting opinions arise, but I'll just leave my feedback here.

RAnders00:
We have found that it would be benefitial to include the CloudFlare domain stuff anyways - at least the certbot how-to from the old steps: https://github.com/pajbot/pajbot/blob/d15b88a9dde00a498157ffa046e05d14259ca53f/install-docs/debian10/install-debian10.txt#L99-L120

alazymeme:
@RAnders00 Will that include the nginx instructions, or is that apart of the new docs?

RAnders00:
the new docs should be clearer about the fact you're supposed to put the streamer's name into every place it says `streamer_name`, especially in the example config file.

I should also put the "full" nginx setup stuff back, sort of as an optional "If you need a full nginx setup to get you started" thing.

alazymeme:
Few things:
- add `sudo apt update` and `sudo apt upgrade` to the start of the script - Just running these two commands fixed some issues for me (a regular pleb wouldn't have thought of this)

- I was running into some permissions issues, but I need to retest this

- The example config file is very "confusing", all the information just seems cluttered with all the comments there. I understand that the comments are beneficial for idiots like me, but they genuinely just confuse me more 4Head

Other than that and adding the certificate instructions from your other guide (for SSL), I also ran into an issue with my own incompetence of not reading the guide. Otherwise it's pretty easy to follow BroBalt

alazymeme:
Might be slightly more useful to include all of these lines in the new guide:

https://github.com/pajbot/pajbot/blob/d15b88a9dde00a498157ffa046e05d14259ca53f/install-docs/debian10/install-debian10.txt#L76-L120

alazymeme:
Also one more thing; add an https redirect config somewhere in the tutorial

alazymeme:
Will run through this again when I get home

alazymeme:
Will test again once the docs get updated with the fixes pajaDank - ping me when they get updated, I've got a channel I need to add it to

alazymeme:
Don't forget to add the http to https redirect

PR #634, opened 2019-11-03
Weather Module

alazymeme:
This is my scuffed attempt at making a weather module using the darksky api. This was inspired by supibot and while existing simple apis exist (like api.scorpstuff.com), this was more me trying to learn more about python by making a simple module - in turn it also helps me grasp the concept of an "api". I'll be using other modules as inspiration, or "learning tool" perse; as it seems the only way I learn is by actually doing the action. This has proved through school and work. Once I start trying to do the action, I should learn relatively quickly.

Over the next few weeks, I'll try and work on this and be done. I've added as a draft pull request in case people want to monitor progress over the development period.


alazymeme:
I'm starting to add a basic units option to the module options so one can switch between celsius and farenheit. Not sure why anyone would use the latter...
Need to brainstorm some more ideas of what config options to add. I don't see there being many however. Perhaps a default location if no argument is defined (of the streamer) - this could be a very broad location or a very precise one.

alazymeme:
I'm still figuring out what needs to be imported into the module in order for it to function correctly. So far I've added some basic settings to the module; I need to figure out the easiest way to list languages in order for the code to be correct - possibly a dropdown or something. I've also fixed up some errors and added a few more module settings. The location option might be a little bit difficult; my thoughts are that I need to use Google Maps API to convert location to lat/long (as those are the options that the Dark Sky API take).

On top of this, I've imported the dark sky api key from the config file and need to parse query parameters for the api. I don't see this being too difficult.

alazymeme:
I think I've nearly finished the module. Need to test and modify accordingly. Once I've confirmed it's all good, I need to modify the module to parse the location through Google's Geo API and parse the lat/long results to the darksky api. Currently the weather module will only support lat/long input.

alazymeme:
You can see there are a few commits above. Mostly due to trial and error.

alazymeme:
If anyone'd like to help me, I'm sort of stuck at this point if you're wiling to take a look Okayga

RAnders00:
What part are you stuck on?

alazymeme:
I get a cb error when trying to start up my bot

alazymeme:
@RAnders00 https://gist.githubusercontent.com/alazymeme/5be175103eb3bb21912e86e904a6dec2/raw/43142cf17341d06d0075f5bebfcc853547c27b2e/v1HxQtrje5.txt

templates/contact.html
```diff
@@ -6,6 +6,6 @@
 <h2>Contact</h2>
 <p>Hi, I'm pajlada! I'm the current maintainer and developer of {{ bot.name }}.</p>
 <p>{{ bot.name }} is running on the <strong>pajbot</strong> source, which is coded in Python.</p>
-<p>Got any questions or comments about the bot or the website? <a href="http://twitch.tv/message/compose?to=pajlada">Let me know by sending me a twitch message!</a></p>
+<p>Got any questions or comments about the bot or the website? <a href="https://discord.gg/qbRE8WR">You can join discord for support!</a> Or just message me: pajlada#2107</p>
```
RAnders00: Uhh this kind of mixing-in of changes with other stuff in the PR is not exactly ideal

alazymeme: PepeLaugh

RAnders00: This is why it's wiser to do your work on a branch different then master, just as a tip for next time

alazymeme: Yeah I found that out soon afterwards PepeLaugh

PR #660, opened 2019-11-08
Issue explicit VACUUM after user rank table refresh

RAnders00:
Please test with care, I don't know if this works or solves anything.
Also, we should consider setting autovacuum to false on user_rank if we are going to do this, or tweak some autovacuum settings?

Pull request checklist:


RAnders00:
I just discovered it but we could also use `DBManager.engine.execute(text("VACUUM user_rank").execution_options(autocommit=True))`

pajlada:
Currently being tested on scadrial on a bunch of bots, will review tomorrow with the results

Issue #688, opened 2019-11-18
Create $(tb:bot_domain) variable


Felanbird:
I think this would be extremely useful as it would allow for virtually every single command that directs to the webui to be:
```
https://$(tb:broadcaster).$(tb:bot_domain)/*
```

pajlada:
bot_domain should return the entire domain value as specified in the `[web]` section in the config file, so for pajbot it would be `paj.pajlada.se`, for snusbot it would be `forsen.tv`, for nymn it would be `nymn.pajbot.com`. so the commands would look like this: `https://$(tb:bot_domain)/...`

alazymeme:
https://github.com/pajbot/pajbot/pull/693

RAnders00:
Closed via #639

PR #689, opened 2019-11-18
updated install-docs/readme.md further-steps

Felanbird:
~~**don't approve until #688 is completed**~~

added new commands + aliases
~~removed --reply~~
~~changed $(source:name) to $(source:username_raw)~~
~~added $(tb:broadcaster)~~
added $(tb:bot_domain)

Pull request checklist:
- [X] `CHANGELOG.md` was updated, if applicable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unversioned
+
 - Minor: Updated install-docs/readme.md further-steps
 
 ## v1.40
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unversioned
+- Minor: Updated install-docs/readme.md further-steps
 
 ## v1.40
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unversioned
 
 - Minor: Updated install-docs/readme.md further-steps
+- Minor: Added `bot_domain` variable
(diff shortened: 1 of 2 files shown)


RAnders00:
- `$(source:name)` is equivalent to `$(source:username_raw)` - they are both the display name.
- Why remove `--reply`? Reply means if the command is whispered e.g. by mods/the bot owner the response will be a whisper as well instead of a "say" to chat.

Apart from that, the other points are good improvements that I'll happily accept! 🚀

Felanbird:
> * `$(source:name)` is equivalent to `$(source:username_raw)` - they are both the display name.
> * Why remove `--reply`? Reply means if the command is whispered e.g. by mods/the bot owner the response will be a whisper as well instead of a "say" to chat.
> 

I was mistaken on the part of --reply which is why I did already re-include it. ebfe1a5

As for $(source:name) vs $(source:username_raw) I was under the impression that $(source:username_raw) took capitalization into consideration and $(source:name) did not, which I have realized is not true so I will revert. 👍 
image

Felanbird:
despite what github thinks the [changelog is correct](https://github.com/Felanbird/pajbot/blob/installdocs-readme-changes/CHANGELOG.md)
probably the 🤠 way I did it though

RAnders00:
username_raw has become name, and username has become login, because that‘s what twitch calls it these days. https://github.com/pajbot/pajbot/blob/1c500fd88bbbd9c3f22986ba1df16a0af9af294c/docs/variables.txt#L115

PR #696, opened 2019-11-18
Removed the unnecessary line

tolekk:
Line caused nginx to not work on port 80
Adding a # in front of it made it work on both port 80 and 443


RAnders00:
Thank you, I have already taken care of this via pull request #695 - in here: https://github.com/pajbot/pajbot/pull/695/commits/d49d2c81ad6329b8f79702668cff5c2c0d3358ff

pajlada:
Fixed in #695 - thank you

PR #705, opened 2019-12-03
Add the ability to change the bot's prefix

alazymeme:
Pull request checklist:


alazymeme:
Yes I fucked up, accidentally used the `fix-pleblist` branch instead of the correct one

alazymeme:
If it tells me to reformat again, I'm gonna kill someone

pajbot/dispatch.py
```diff
@@ -32,7 +31,7 @@ def add_command(bot, source, message, event, args):
         # Make sure we got both an alias and a response
         message_parts = message.split()
         if len(message_parts) < 2:
-            bot.whisper(source, "Usage: !add command ALIAS [options] RESPONSE")
+            bot.whisper(source, f"Usage: {self.prefix}add command ALIAS [options] RESPONSE")
```
RAnders00: Should be bot.prefix (and in several other places too)

pajbot/dispatch.py
```diff
@@ -19,8 +19,7 @@ def add_win(bot, source, message, event, args):
         bot.kvi["br_wins"].inc()
         bot.me(f"{source} added a BR win!")
 
-    @staticmethod
-    def add_command(bot, source, message, event, args):
+    def add_command(self, bot, source, message, event, args):
```
RAnders00: Add @staticmethod back since self is not needed

pajbot/managers/command.py
```diff
@@ -115,7 +115,7 @@ def load_internal_commands(self):
                 CommandExample(
                     None,
                     "Default usage",
-                    chat="user:!twitterfollow forsensc2\n" "bot>user:Now following ForsenSC2",
+                    chat=f"user:{self.prefix}twitterfollow forsensc2\n" "bot>user:Now following ForsenSC2",
```
RAnders00: Same here, you need bot.prefix

alazymeme:
@RAnders00 what should I do here: https://github.com/alazymeme/pajbot/blob/master/pajbot/web/routes/api/commands.py#L161 in order to "import" the prefix (using the bot definition or whatever it's called fdm)

RAnders00:
```python
from flask import current_app as app

log_msg = f'The {app.bot_prefix}{command.command.split("|")[0]} command has been updated from "{old_message}" to "{new_message}"'
```

And in `web/__init__.py`, add a config parsing entry like you added in `bot.py`, this time call it `app.bot_prefix = ...` similar to how the other existing ones are already done:

https://github.com/pajbot/pajbot/blob/master/pajbot/web/__init__.py#L115

alazymeme:
@RAnders00 Check the above commit, is this required in every file that I cannot "import" bot?

RAnders00:
In module files the bot is available via `self.bot`.

The only pöaces you need to do it through `app` is in the web process which is in the .py files in the `web` directory.

alazymeme:
@pajlada You haven't commented on this, so I just have to tag you and ask for your review

PR #715, opened 2019-12-10
Fixed sub emotes

TroyDota:
Api for getting sub emotes from the https://twitchemotes.com/

Pull request checklist:


alazymeme:
Correct me if I'm wrong, but how do you suggest that simply adding a new python file will fix things? Should it not be referenced somewhere?

TroyDota:
Yes but I fucked the pull so there is one other file that must change, the
emote manager file. I'll do it now

PR #758, opened 2020-01-17
Add emotes to welcome/leave messages

alazymeme:


alazymeme:
The commits below "Update example.ini` are all old commits. No idea why they're there, but they've already been merged with master

RAnders00:
I suggest "HeyGuys" or "SeemsGood" and BibleThump maybe because they are native twitch emotes?

PR #791, opened 2020-02-20
Added a ban reason for "!permaban"

Felanbird:
Added a generic ban reason for !permaban as it had no ban reason before which could cause confusion as it could be seen as the bot "randomly" banning if the !permaban usage was forgotten.

Also I included the level section in the ban reason only because the user does not receive the ban reason so it would not cause any confusion for them, just as a note for the mods.

image

Pull request checklist:

- [X] `CHANGELOG.md` was updated, if applicable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Added ban reason for "!permaban"
+
 ## v1.42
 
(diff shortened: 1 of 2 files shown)


Felanbird:
If a maintainer could fix my formatting that would be appreciated, windows own instance of python is blocking my permissions and preventing me from running the script 😤
(I believe)

PR #823, opened 2020-04-05
Fix founder badge logic

pajlada:
When a user has the founder badge, the "subscriber" tag is surprisingly
set to 0. We previously relied on the "subscriber" tag, but now instead
check whether the user has a subscriber or founder badge.

Pull request checklist:

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -643,7 +643,7 @@ class Bot:
         emote_tag = tags["emotes"]
         msg_id = tags.get("id", None)  # None on whispers!
         badges_string = tags.get("badges", "")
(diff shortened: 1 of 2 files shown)


pajbot/bot.py
```diff
@@ -642,11 +642,14 @@ def parse_message(self, message, source, event, tags={}, whisper=False):
 
         emote_tag = tags["emotes"]
         msg_id = tags.get("id", None)  # None on whispers!
+        badges_string = tags.get("badges", "")
+        badges = dict((badge.split("/") for badge in filter(None, badges_string.split(","))))
```
RAnders00, line 646:
```suggestion
        badges = dict((badge.split("/") for badge in badges_string.split(",") if badge != ""))
```

Issue #763, opened 2020-01-19
Fix urlfetch censoring on the website for nested substitution cases


alazymeme:
Make it so that http/https isn't required. Currently the urlfetch parameter isn't masked because it doesn't contain http/https: https://pajlada.pajbot.com/commands/1057-shuffle
![image](https://user-images.githubusercontent.com/12804673/72679810-8d015380-3afe-11ea-81d8-24590130103b.png)

RAnders00:
The command actually contains `https://`:

`17:02 pajbot -> randers: id=1057, level=100, type=message, cost=0, cd_all=5, cd_user=15, mod_only=False, response=$(urlfetch https://shitty-customapis-si5z3vz69j2s.runkit.sh/random/shuffle?input=$(args:0|urlencode))`

The `$(args:0|urlencode)` does not match the urlfetch regex. I suspect it normally (in the bot) gets replaced before urlfetch matches are performed.

RAnders00:
I think this is more of a feature request rather than a bug. For this reason due to pajbot [being moved into maintenance mode](https://github.com/pajbot/pajbot/blob/master/README.md) I'm closing this.

alazymeme:
I disagree, this is a bug. It's an implemented feature that isn't working.

RAnders00:
Unless you‘re already working on a fix because this is arguably not a major breakage in any way I didnt find it to be worth to keep the open bug lingering around. I‘ll assign you and reopen

alazymeme:
This will be closed at some point anyway. Might as well close it now.

PR #825, opened 2020-04-05
Fix urlfetch masking

alazymeme:
Pull request checklist:

Fixes https://github.com/pajbot/pajbot/issues/763

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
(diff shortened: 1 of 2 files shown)


pajlada:
Not the way I want this to be fixed.

To clarify, the issue comes from the website not parsing out "inner arguments" in the urlfetch string, while the bot does. So any string that can occur in an "inner argument" (e.g. `$(args:0|urlencode)`) can ruin the urlfetch masking.

We could have a separate regex that is looser with its matching, and use that as the website masking regex

PR #830, opened 2020-04-13
Lowered runpnsl minimum level to 420

Felanbird:
I didn't update the changelog because python isn't my friend, if this is an issue I apologize and it will have to be added by a maintainer 😢 

Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Minor: Lowered minimum level required for the `!runpnsl` command from 500 to 420.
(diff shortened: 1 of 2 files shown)


RAnders00:
one second please while i take a look too

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Lowered minimum level required for the `!runpnsl` command from 500 to 420.
```
RAnders00, line 5:
```suggestion
- Minor: You can now allow users with level 420 to use the `!runpnsl` command (could only be set as low as 500 before). Default level requirement remains at 750. (#830)
```

PR #836, opened 2020-04-19
Emote Limit Module

### LINKED ISSUE #834: Add an emote spam filter
Timeout messages containing excessive amounts of emotes.
If max is set at 15 and the user sends 18 emotes in a message it times them out for a chosen amount of time. This would be able to check for both regular Twitch emotes and 3rd party emotes.

pajlada:
simple module that can limit number of emotes users can post in chat

image

Fixes #834 

Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added new module for simple emote spam moderation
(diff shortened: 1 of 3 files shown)


alazymeme:
Does this work for emojis also? Or is that apart of another module?

RAnders00:
@alazymeme this won't work for emojis.

PR #860, opened 2020-05-09
Fix login flow

pajlada:
The login flow didn't work because we made a "raw" api request that didn't contain any of the now-required headers.

This PR aims to fix this :slightly_smiling_face: 

Pull request checklist:

### DIFF
--- a/pajbot/apiwrappers/twitch/helix.py
+++ b/pajbot/apiwrappers/twitch/helix.py
@@ -233,3 +233,13 @@ class TwitchHelixAPI(BaseTwitchAPI):
             else None
             for user_data in bulk_user_data
         ]
+
+    def get_user_data_by_client_id_and_access_token(self, client_id, access_token):
+        response = self.get(
+            "/users", headers={"Authorization": "Bearer " + access_token, "Client-ID": client_id}, authorization={}
+        )
+
+        if len(response["data"]) <= 0:
+            return None
+
+        return response["data"][0]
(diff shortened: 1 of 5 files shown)


pajbot/web/__init__.py
```diff
@@ -60,6 +60,9 @@ def init(args):
     twitch_helix_api = TwitchHelixAPI(RedisManager.get(), app_token_manager)
     twitch_badges_api = TwitchBadgesAPI(RedisManager.get())
 
+    app.twitch_helix_api = twitch_helix_api
+    app.twitch_id_api = id_api
```
RAnders00, line 64: should rename the local variable to `twitch_id_api`

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -233,3 +233,13 @@ def bulk_get_user_basics_by_login(self, logins):
             else None
             for user_data in bulk_user_data
         ]
+
+    def get_user_data_by_client_id_and_access_token(self, client_id, access_token):
```
RAnders00, line 237: method name should start with `fetch` instead of `get` since it never uses a cache of any sort, in line with how the rest of the methods in this api wrapper are named (e.g. `fetch_all_subscribers`)

pajbot/web/routes/base/login.py
```diff
@@ -104,25 +104,24 @@ def authorized():
...
 
         with DBManager.create_session_scope(expire_on_commit=False) as db_session:
             me = User.from_basics(
-                db_session,
-                UserBasics(
-                    me_api_response.data["data"][0]["id"],
-                    me_api_response.data["data"][0]["login"],
-                    me_api_response.data["data"][0]["display_name"],
-                ),
+                db_session, UserBasics(me_api_response["id"], me_api_response["login"], me_api_response["display_name"])
```
RAnders00, line 117: to avoid this, the helix API wrapper could get a method that returns UserBasics given some authorization

(but I would see it as optional)

Issue #450, opened 2019-08-11
API wrapper refactor: Try use streamer authorization for title/game updates


RAnders00:
- request the requisite scope in `streamer_login`
- on auth fail with the bot's authorization, retry with the streamer's.
- If both fail: `bot.say` error message like right now.

PR #861, opened 2020-05-09
Reworked login system

### LINKED ISSUE #816: flask-oauthlib has been deprecated
This causes issues with the recent release of werkzeug 1.0.0 that breaks this library.

[authlib](https://github.com/lepture/authlib) is what the developer is recommending everyone to migrate to.

RAnders00:
Login is now (imo) nicer and more maintainable

- Remove dependency on deprecated flask-oauthlib (Closes #816)
- added csrf validation to login process
- Makes login code compatible with upcoming changes to helix authentication
- Login error page now tells you what went wrong
- We don't store the twitch access token in the session cookie for no reason anymore
- We no longer request the scope to read the user's email by default, the app works just fine with an empty list of scopes.

Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Minor: Updated semantic-ui css and js dependencies
 - Minor: Login system is now protected against CSRF attacks (#861)
 - Minor: Updated code for login system to make it work with upcoming changes to the Twitch API. (#861)
+- Minor: Login process no longer asks users for the permission to read the email from their profile (we now only request public information). (#861)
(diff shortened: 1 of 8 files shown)


alazymeme:
Any chance you'd want to have a look at this as well? :D https://github.com/pajbot/pajbot/issues/450
Since it's something to do with the login that you're fixing.

RAnders00:
That's not really login related, and does not fit into this PR at all. We can do it afterwards, and it would be just as easy to do as before this PR.

PR #872, opened 2020-05-16
Reconnect on connection error

pajlada:
Previously if a connection error handled during the actual "connection phase" (e.g. connection timed out, connection refused), we ignored it and never tried to connect again.

This attempts to remedy that, although it's not something I have set up a test case for but it worked for Leppunen once babyrage

Pull request checklist:

### DIFF
--- a/pajbot/managers/connection.py
+++ b/pajbot/managers/connection.py
@@ -91,7 +91,7 @@ class ConnectionManager:
 
             return True
         except:
-            log.exception("Connection failed, retrying")
+            log.exception("Failed to open connection, retrying")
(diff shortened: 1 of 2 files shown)


pajbot/managers/connection.py
```diff
@@ -91,24 +91,22 @@ def start(self):
 
             return True
         except:
-            log.exception("babyrage")
+            log.exception("Connection failed, retrying")
```
RAnders00, line 94:
```suggestion
            log.exception("Failed to open connection, retrying")
```

"Connection failed" to me sounds like we had a connection, but it failed later

PR #878, opened 2020-05-18
TMI Rates

### LINKED ISSUE #691: verified = 1 flag should do more
If the option isn't defined, I believe the bot should be far more limited. Referencing this guide: https://dev.twitch.tv/docs/irc/guide

If the `verified = 1` option is not in the config:
- Bots should be limited to 100 messages/30 seconds (assuming moderator status)
- Whispers should be limited to 3/second; 100/minute
- Eval command should also be limited (using eval command for bans, etc)

This will definitely help. I was importing playsounds for a mate; his bot was rate limited as a result of trying to spam whispers. If this is the case, then perhaps the bot should collate it's whispers?

`Successfully imported x playsounds` etc

TroyDota:
Pull request checklist:

Fixes #691

### DIFF
--- a/configs/example.ini
+++ b/configs/example.ini
@@ -25,7 +25,8 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; if your bot is known set known to 1
 ;known = 1
 ; if you wish to change how whipers are managed. setting whisper_output to 0 will disable whispers, 
-; setting it to 2 will output all whsipers in chat. The default setting 1 is whispers act as whispers.
+; setting it to 2 will output all whsipers in chat. The default setting 1 is to rate-limit whispers according to
+; the verified = 1 or known = 1 settings.
 ;whisper_output = 1
 
 ; Optional section if you want to make the "Wolfram Alpha Query" module available for use:
--- a/configs/example.ini
+++ b/configs/example.ini
@@ -24,10 +24,9 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ;verified = 1
 ; if your bot is known set known to 1
 ;known = 1
-; if you wish to change how whipers are managed. setting whisper_output to 0 will disable whispers, 
-; setting it to 2 will output all whsipers in chat. The default setting 1 is to rate-limit whispers according to
-; the verified = 1 or known = 1 settings.
-;whisper_output = 1
+; if you wish to change how whipers are managed. setting whispers to "disabled" will disable whispers, 
+; setting it to "chat" will output all whsipers in chat. The default setting "normal" is whispers act as normal whispers.
+;whispers = normal
 
 ; Optional section if you want to make the "Wolfram Alpha Query" module available for use:
 ; Set this to a valid Wolfram|Alpha App ID to enable wolfram alpha query functionality
--- a/configs/example.ini
+++ b/configs/example.ini
@@ -25,10 +25,10 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; if your bot is known (increased rate limits, see same link above) set this to 1
 ;known = 1
 ; Allows you to set a different output mode for whispers.
-; `whispers = normal` sends whispers as normal. (This is the default if you don't configure this option)
-; `whispers = disabled` will make it so the bot does not send any whispers. Outgoing whispers will be dropped.
-; `whispers = chat` outputs all messages intended to be whispered to the streamer's chat instead of whispering them.
-;whispers = normal
+; `whisper_output_mode = normal` sends whispers as normal. (This is the default if you don't configure this option)
+; `whisper_output_mode = disabled` will make it so the bot does not send any whispers. Outgoing whispers will be dropped.
+; `whisper_output_mode = chat` outputs all messages intended to be whispered to the streamer's chat instead of whispering them.
+;whisper_output_mode = normal
 
 ; Optional section if you want to make the "Wolfram Alpha Query" module available for use:
 ; Set this to a valid Wolfram|Alpha App ID to enable wolfram alpha query functionality
--- a/configs/example.ini
+++ b/configs/example.ini
@@ -25,7 +25,7 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; if your bot is known set known to 1
 ;known = 1
 ; if you wish to change how whipers are managed. setting whispers to "disabled" will disable whispers, 
-; setting it to "chat" will output all whsipers in chat. The default setting "normal" is whispers act as normal whispers.
+; setting it to "chat" will output all whsipers in the streamers chat, where as setting this to "disabled" will stop the bot from trying to whisper entirely. The default setting "normal" makes no changes to the way whispers act. 
 ;whispers = normal
 
 ; Optional section if you want to make the "Wolfram Alpha Query" module available for use:
--- a/configs/example.ini
+++ b/configs/example.ini
@@ -22,10 +22,12 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; set this to 1 if your bot is a verified bot (increased rate limits) on Twitch
 ; More info about verified bots can be found here: https://dev.twitch.tv/docs/irc/guide#known-and-verified-bots
 ;verified = 1
-; if your bot is known set known to 1
+; if your bot is known (increased rate limits, see same link above) set this to 1
 ;known = 1
-; if you wish to change how whipers are managed. setting whispers to "disabled" will disable whispers, 
-; setting it to "chat" will output all whsipers in the streamers chat, where as setting this to "disabled" will stop the bot from trying to whisper entirely. The default setting "normal" makes no changes to the way whispers act. 
+; Allows you to set a different output mode for whispers.
+; `whispers = normal` sends whispers as normal. (This is the default if you don't configure this option)
+; `whispers = disabled` will make it so the bot does not send any whispers. Outgoing whispers will be dropped.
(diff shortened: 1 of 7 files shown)


TroyDota:
@alazymeme had the known bot idea

configs/example.ini
```diff
@@ -22,6 +22,11 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; set this to 1 if your bot is a verified bot (increased rate limits) on Twitch
 ; More info about verified bots can be found here: https://dev.twitch.tv/docs/irc/guide#known-and-verified-bots
 ;verified = 1
+; if your bot is known set known to 1
+;known = 1
+; if you wish to change how whipers are managed. setting whisper_output to 0 will disable whispers, 
+; setting it to 2 will output all whsipers in chat. The default setting 1 is whispers act as whispers.
```
RAnders00, line 28: this needs to be explained better, e.g.:

```suggestion
; if you wish to change how whipers are managed. setting whisper_output to 0 will disable whispers, 
; setting it to 2 will output all whsipers in chat. The default setting 1 is to rate-limit whispers according to
; the verified = 1 or known = 1 settings.
```

pajbot/bot.py
```diff
@@ -88,6 +88,12 @@ def __init__(self, config, args):
 
         if config["main"].getboolean("verified", False):
             TMI.promote_to_verified()
+        elif config["main"].getboolean("known", False):
+            TMI.promote_to_known()
```
RAnders00, line 92: swap these, so `known = 1` and `verified = 1` lead to rate limits of a verified bot

```suggestion
        if config["main"].getboolean("known", False):
            TMI.promote_to_known()
        elif config["main"].getboolean("verified", False):
            TMI.promote_to_verified()
```

configs/example.ini
```diff
@@ -22,6 +22,11 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
...
+;whisper_output = 1
```
RAnders00, line 29: I'd prefer if this was a string-based setting, e.g. `whispers = disabled/ratelimited/unlimited`

pajbot/bot.py
```diff
@@ -88,6 +88,12 @@ def __init__(self, config, args):
...
+
+        whisper_output = config["main"].getint("wisper_output", 1).lower()
+        if whisper_output in [0, 2]:
+            TMI.change_whisper_output(whisper_output)
```
RAnders00, line 96: I don't like it being called "whisper output". Can we call it "whisper limiting" or "whisper rate-limiting" or something like that?

pajbot/managers/connection.py
```diff
@@ -51,13 +51,26 @@ def __init__(self, reactor):
         super().__init__(reactor)
 
         self.num_msgs_sent = 0
+        self.num_whispers_sent_minuite = 0
```
RAnders00, line 54: typo

pajbot/managers/connection.py
```diff
@@ -51,13 +51,26 @@ def __init__(self, reactor):
...
+        self.num_whispers_sent_second = 0
         self.in_channel = False
 
     def reduce_msgs_sent(self):
         self.num_msgs_sent -= 1
 
-    def can_send(self):
-        return self.num_msgs_sent < TMI.message_limit
+    def reduce_whispers_sent_minuite(self):
```
RAnders00, line 61: typo

pajbot/managers/connection.py
```diff
@@ -51,13 +51,26 @@ def __init__(self, reactor):
...
+        self.num_whispers_sent_minuite -= 1
+
+    def reduce_whispers_sent_second(self):
```
RAnders00, line 64: these defs should be prefixed with _ since they're private
The fact that previous pajlada or randers didn't do that shouldn't prevent you from following best practices :)

pajbot/managers/connection.py
```diff
@@ -51,13 +51,26 @@ def __init__(self, reactor):
...
+        self.num_whispers_sent_second -= 1
+
+    def can_send(self, whisper=False):
+        if whisper:
+            return self.num_msgs_sent < TMI.message_limit
+        return (
+            self.num_whispers_sent_second < TMI.whispers_message_limit_second
+            and self.num_whispers_sent_minuite < TMI.whispers_message_limit_minute
```
RAnders00, line 72: typo

pajbot/managers/connection.py
```diff
@@ -112,15 +125,25 @@ def on_disconnect(self, _chatconn):
         log.error("Disconnected from IRC")
         self.start()
 
-    def privmsg(self, channel, message, increase_message=True):
+    def privmsg(self, channel, message, increase_message=True, whisper=False):
```
RAnders00, line 128: I would prefer if you didn't introduce a `whisper` parameter and instead did the limiting logic in the `whisper` method

pajbot/managers/connection.py
RAnders00: Additionally, when sending a whisper, num_msgs_sent also counts (the 7500 or 100 are still counted when sending whispers, because they are for all PRIVMSG messages)

pajbot/managers/connection.py
```diff
@@ -112,15 +125,25 @@ def on_disconnect(self, _chatconn):
...
         conn = self.main_conn
 
-        if conn is None or not conn.can_send():
+        if TMI.disable_whisper and whisper:
+            log.warning("Whispers disabled in config yet, whisper requested")
```
RAnders00, line 132:
```suggestion
            log.debug("Whisper was not sent (due to config setting)")
```

pajbot/tmi.py
```diff
@@ -1,8 +1,20 @@
 class TMI:
```
RAnders00, line 1: you should introduce a new enum class (https://docs.python.org/3/library/enum.html) that has the three variants for whisper limits. Preferrably add the string parsing for the config value to the enum class (e.g. `def from_config_value(config_value)`)

pajbot/managers/connection.py
```diff
@@ -112,15 +125,25 @@ def on_disconnect(self, _chatconn):
...
+
+        if conn is None or not conn.can_send(whisper):
             log.error("No available connections to send messages from. Delaying message a few seconds.")
```
RAnders00, line 135: fix this message while you're at it

pajbot/managers/connection.py
TroyDota: How do we know if the message is a whisper?

pajbot/managers/connection.py
TroyDota: They cant be private because they are apart of the Connection class but are referenced in the ConnectionManager class.

RAnders00: Ah I see, my bad.

TroyDota:
Is there more i must fix/add?

RAnders00:
I think I want @pajlada to review this and work on it

TroyDota:
Sure.

pajbot/bot.py
RAnders00: This is still a problem, verified should take precedence over known

RAnders00: I overlooked the elif you introduced, my bad

TroyDota:
OkayChamp I'll make the entry

TroyDota:
i did these commits via the web interface WAYTOODANK

TroyDota:
Is there more to change for this?

pajbot/tmi.py
```diff
@@ -1,8 +1,33 @@
...
+    NORMAL = 1
+    CHAT = 2
+
+    @staticmethod
+    def from_config_value(config_value):
+        return {
+            "disabled": WhisperOutputMode.DISABLED,
+            "normal": WhisperOutputMode.NORMAL,
+            "chat": WhisperOutputMode.CHAT,
+        }.get(config_value.lower(), WhisperOutputMode.NORMAL)
```
RAnders00, line 15: Last discussion: Is this behaviour good? Should an invalid config option (e.g. a missspelling) be silently swallowed?

TroyDota: no, i think an invalid config option should throw a warning and then revert to normal behavior.

RAnders00, line 15:
```suggestion
        }.get(config_value.lower())
```

I'm suggesting it should just crash, that's easy to implement.

RAnders00, line 15:
```suggestion
    CONFIG_OPTIONS = {
            "disabled": WhisperOutputMode.DISABLED,
            "normal": WhisperOutputMode.NORMAL,
            "chat": WhisperOutputMode.CHAT,
        }

    @staticmethod
    def from_config_value(config_value):
        if config_value.lower() in CONFIG_OPTIONS:
            return CONFIG_OPTIONS[config_value.lower()]
        else:
            raise ValueError("whisper_output_mode config option is not valid.")
```

pajlada: I agree with RAnders, since we already grab a default value "normal" if no value is set, adding a crash on misconfiguration seems fine. Adding the crash adds no extra friction for already created bots

TroyDota: 👍

configs/example.ini
```diff
@@ -22,6 +22,13 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; set this to 1 if your bot is a verified bot (increased rate limits) on Twitch
 ; More info about verified bots can be found here: https://dev.twitch.tv/docs/irc/guide#known-and-verified-bots
 ;verified = 1
+; if your bot is known (increased rate limits, see same link above) set this to 1
+;known = 1
+; Allows you to set a different output mode for whispers.
+; `whispers = normal` sends whispers as normal. (This is the default if you don't configure this option)
+; `whispers = disabled` will make it so the bot does not send any whispers. Outgoing whispers will be dropped.
+; `whispers = chat` outputs all messages intended to be whispered to the streamer's chat instead of whispering them.
+;whispers = normal
```
RAnders00, line 31: Should this be called `whisper_output_mode`

PR #882, opened 2020-05-19
Clear Chat

### LINKED ISSUE #881: Clear Chat function problems
- Throws an error when /clear command is called from chat
- Permaban doesnt do anything in the database.

TroyDota:
Pull request checklist:

Fixes #881

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -802,6 +802,7 @@ class Bot:
     def on_clearchat(self, chatconn, event):
         tags = {tag["key"]: tag["value"] if tag["value"] is not None else "" for tag in event.tags}
 
+        # Ignore "Chat has been cleared by a moderator" messages
         if "target-user-id" not in tags:
             return
 
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -802,6 +802,9 @@ class Bot:
     def on_clearchat(self, chatconn, event):
         tags = {tag["key"]: tag["value"] if tag["value"] is not None else "" for tag in event.tags}
 
+        if "target-user-id" not in tags:
+            return
+
         target_user_id = tags["target-user-id"]
         with DBManager.create_session_scope() as db_session:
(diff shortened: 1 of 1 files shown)


pajbot/bot.py
```diff
@@ -819,7 +822,8 @@ def on_clearchat(self, chatconn, event):
             else:
                 # permaban
                 # this sets timeout_end to None
-                user.timed_out = False
+                user.timed_out = True
```
RAnders00, line 825: You can't set `timed_out` to true:

https://github.com/pajbot/pajbot/blob/37da06f4f18237486513e86eb2efcb80d5755bf4/pajbot/models/user.py#L173-L178

The only data we store for each user is a `timeout_end`, as you can imagine, you can't fit permabans into that sort of model, which is why we just clear the timeout_end field on permaban.
Reason is: There's just currently no use to knowing a user is banned (so why bother implementing it/storing it?). the only thing we use `timeout_end` is the paid timeout module (so you can stack timeouts).

pajbot/bot.py
```diff
@@ -802,6 +802,9 @@ def on_pubnotice(self, chatconn, event):
     def on_clearchat(self, chatconn, event):
         tags = {tag["key"]: tag["value"] if tag["value"] is not None else "" for tag in event.tags}
 
+        if "target-user-id" not in tags:
```
RAnders00, line 805: Please add a comment:

```suggestion
        # Ignore "Chat has been cleared by a moderator" messages
        if "target-user-id" not in tags:
```

pajbot/bot.py
```diff
@@ -817,7 +821,8 @@ def on_clearchat(self, chatconn, event):
             else:
                 # permaban
                 # this sets timeout_end to None
-                user.timed_out = False
+                user.timed_out = True
+                user.timeout_end = None
```
RAnders00, line 825:
```suggestion
                user.timed_out = False
```

Issue #671, opened 2019-11-10
Show vip badge on profile


alazymeme:
https://i.nuuls.com/7E0qd.png

alazymeme:
👎🏿

alazymeme:
Fixed by https://github.com/pajbot/pajbot/pull/886

PR #886, opened 2020-05-20
Badges Recording

TroyDota:
Pull request checklist:


TroyDota:
This allows for the modules to use the fact that they are "founders" or that they are "vips" allowing for, when i add my songrequest module vip/founder only requests 💯

pajbot/migration_revisions/db/0012_badges.py
```diff
@@ -0,0 +1,3 @@
+def up(cursor, bot):
+    cursor.execute('ALTER TABLE "user" ADD COLUMN IF NOT EXISTS founder BOOLEAN NOT NULL DEFAULT FALSE')
+    cursor.execute('ALTER TABLE "user" ADD COLUMN IF NOT EXISTS vip BOOLEAN NOT NULL DEFAULT FALSE')
```
RAnders00, line 3: why is there a `IF NOT EXISTS` in there, shouldn't the schema upgrade fail if the column already exists?

pajbot/models/user.py
```diff
@@ -106,6 +108,8 @@ def __init__(self, *args, **kwargs):
         self.ignored = False
         self.banned = False
         self.timeout_end = None
+        self.vip = False
+        self.founder = False
 
         super().__init__(*args, **kwargs)
```
RAnders00, line 114: please add the new properties to the `jsonify` method

pajbot/migration_revisions/db/0012_badges.py
TroyDota: oh i left it in cause i compressed a bunch of schema changes, ill remove it by bad.

TroyDota:
Web interface WAYTOODANK

TroyDota:
> Additionally, I would expect a change to the web interface to show VIP and founder badges if a user has these.
> 
> Also, you should extend the moderators refresh module to also refresh VIPs.

Ill do this soon sorry for being inactive :P

TroyDota:
@RAnders00 What do u think? @alazymeme did it.

alazymeme:
> @RAnders00 What do u think? @alazymeme did it.

idk if what I did needs more, but I thought I'd try and reboot the pr

alazymeme:
> 
> * extension to the moderator refresh to also refresh VIPs

Do you specifically want an extension, or a new module? 🤔

RAnders00:
Would be nice to refactor the moderators refresh module into one that does both

alazymeme:
> Would be nice to refactor the moderators refresh module into one that does both, to prevent code duplication

Will renaming it screw anything up in the db?

TroyDota:
> > Would be nice to refactor the moderators refresh module into one that does both, to prevent code duplication
> 
> Will renaming it screw anything up in the db?

No, renaming it will create a new module state, so every bot will have an extra entry in the database but who cares.

TroyDota:
ANy news on this pr?

RAnders00:
@TroyDota see my review, it still applies. I will not merge this as is.

alazymeme:
Fixes https://github.com/pajbot/pajbot/issues/671

TroyDota:
the fuck. git has the master changes on this pr? Can this be resolved or do we need to make a new pr?

alazymeme:
Pls re-review when you get the chance

CHANGELOG.md
```diff
@@ -16,9 +16,8 @@
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
 - Major: Added a new config option `whisper_output_mode`, allowing you to send whispers to the streamer's chat or to disable them altogether, to aid un-verified bots that may be unable to send whispers. View the example config for more info. (#878)
-- Minor: Added an extra config option for _known_ bots (in addition to verified bots that were already supported). See example config for usage. (#878)
-- Minor: Moved chat alerts into a sub-category. (#896)
-- Minor: Added a new module to print a chat alert on raids. (#896)
```
RAnders00, line 21: something was merged incorrectly here

RAnders00, line 21: also, still missing the changelog entry

pajbot/modules/vip_refresh.py
```diff
@@ -0,0 +1,151 @@
...
+        self.bot.privmsg("/vips")
+
+    @staticmethod
+    def _parse_pubnotice_for_vips(msg_id, message):
+        if msg_id == "no_vips":
+            return []
+
+        if msg_id == "vips_success":
+            if message.startswith("The VIPs of this channel are: "):
+                message = message[30:]  # 36 = length of the above "prefix"
```
RAnders00, line 46: mismatch, comment says 36 and code says 30

pajbot/modules/vip_refresh.py
```diff
@@ -0,0 +1,151 @@
...
+                f"Received vips_success NOTICE message, but actual message did not begin with expected prefix. Message was: {message}"
+            )
+
+        return None
+
+    def _on_pubnotice(self, channel, msg_id, message):
+        if channel != self.bot.streamer:
+            return
+
+        vip_logins = self._parse_pubnotice_for_vips(msg_id, message)
```
RAnders00, line 58: something got cut off here. Since the rest is mostly copied from the moderators refresh, check here for what it should look like: https://github.com/pajbot/pajbot/blob/master/pajbot/modules/moderators_refresh.py#L61-L66

pajbot/modules/vip_refresh.py
```diff
@@ -0,0 +1,151 @@
...
+                "vips": Command.raw_command(
+                    self.update_vip_cmd,
+                    delay_all=120,
+                    delay_user=120,
+                    level=1000,
+                    examples=[
+                        CommandExample(
+                            None,
+                            f"Reload who is a Twitch channel VIP",
+                            chat=f"user:!reload vips\nbot>user: Reloading VIP status...",
```
RAnders00, line 125: the command as it is right now actually says `Reloading list of VIPs...`, not `Reloading VIP status...`, one of the two should be corrected to be consistent.
I also realize that this same mistake is also in the moderators refresh module, but that can be fixed separately.

pajbot/modules/vip_refresh.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+
+    def enable(self, bot):
+        # Web interface, nothing to do
+        if not bot:
+            return
+
+        HandlerManager.add_handler("on_pubnotice", self._on_pubnotice)
+
+        # every 10 minutes, send the /mods command (response is received via _on_pubnotice)
```
RAnders00, line 139: forgot to change it from `/mods` to `/vips`

RAnders00:
I should say overall, copying the moderators refresh module like this and just changing the bits that say "moderator" to instead be "vip" is a big anti-pattern and as you can see from the review is proving to be error-prone. A better approach would be to take the existing module and refactor it to do both things, if possible.

(Just a general comment. I plan on accepting this PR as it is once the bugs are ironed out)

CHANGELOG.md
alazymeme: I can't mark as solved, so I'll just put a 👍

alazymeme: 👍

pajbot/modules/vip_refresh.py
alazymeme: 👍

pajbot/modules/vip_refresh.py
alazymeme: Yes - why would the broadcaster need VIP privileges if they have mod privileges? I removed that part and feel that there's more that I need to remove

pajbot/modules/vip_refresh.py
alazymeme: 👍

pajbot/modules/vip_refresh.py
alazymeme: 👍

pajbot/modules/vip_refresh.py
RAnders00: the way you removed it makes it parse the message, and then nothing is done with what has been parsed. You don't call `_process_vip_logins` anywhere inside the module, which is what does the database update

CHANGELOG.md
```diff
@@ -348,4 +349,4 @@ Changelogs were kept sporadically before this point
 
 - Major change to the way we store and cache users. Previously we cached users in-memory for 10 minutes before pushing any changes to the MySQL database. We have no removed this layer entirely, and changes made in the bot will be pushed immediately. This will greatly reduce the memory footprint of pajbot, but put a bigger strain on the database.
 - Moved a bunch of columns from MySQL to redis. (num_lines, last_seen, last_active, banned, ignored)
-- A bunch of more things that I don't remember, first changelog :sunglasses::ok_hand:
+- A bunch of more things that I don't remember, first changelog :sunglasses::ok_hand:
```
RAnders00, line 352: this line probably should not be in the diff, is there some invisible characters or newlines added/removed here?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added VIP support. VIPs will now show in the !debug command and on the userpage on the website. (#886)
```
RAnders00, line 5: Also should add that founders support was added

RAnders00: Additionally, this data is also present on the API, which should be noted

alazymeme: 👍

CHANGELOG.md
alazymeme: 👍

pajbot/modules/vip_refresh.py
alazymeme: Updated 👍

alazymeme:
> i don't want to click approve yet, but I don't see any issues anymore from purely reviewing the code.

I assumed the vip badge not showing was only on my local machine's version. But it seems that it just isn't working on my main bot either. Not sure what it could be though since the space is there for the badge:

image

And the code is there as well:
image

RAnders00:
Now I get it, it's because the SVG badge is white on transparent background. The VIP badge div also needs a background.

static/css/components/badges.sass
```diff
@@ -29,4 +29,10 @@
   &.bot
     background: url('/static/images/badge_bot.png')
     background-size: 100%
+  &.vip
+    background: url('/static/images/badge_vip.png')
+    background-size: 100%
```
RAnders00, line 34:
```suggestion
    background-size: 100%
    background-color: rgb(224, 5, 185)
```

alazymeme:
> Now I get it, it's because the SVG badge is white on transparent background. The VIP badge div also needs a background.

I don't think it's this, since it's still not working. I am noticing that the element itself is missing from the css. Could there be something wrong in the `username_link` file? Since sub and mod work just fine?

image
image

static/css/components/badges.sass
RAnders00: yeah I just realized that this is not actually the case. The image has the background "baked in".

I was using inspect element on the Twitch website. Apparently the FFZ browser extension changes the standard badges to SVG ones with a CSS background.

alazymeme: So should I also do this to be consistent? Or just go with what I have?

alazymeme:
After c97f916, I started getting a new error:
```requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.twitch.tv/helix/users?login=5th_dimension52&login=kraticevil&login=lidl_sub&login=postbach&login=ruk0&login=samtwocan&login=scriptorex&login=seirion&login=slch000&login=william_bonesbot

File: static/css/components/badges.sass (hunk 10 above)
Comment (RAnders00) [reply static/css/components/badges.sass:34]: keep it as it was before

**Comment by pajlada** [comment]
To make this clear:
What's still to be added is:
Badges appearing next to user profiles

**Comment by TroyDota** [comment]
@pajlada it seems, as if the webapp caches the stupid sass files because they were imported and not directly accessed via flask assets. Deleting the cache fixes this change. KINDA A WEIRDCHAMP on flask.

**Comment by TroyDota** [comment]
Also to note, if the bot is a founder the bot WILL NOT have a founder's badge on the website as, the bot ignores its own messages so it never gets to the part there it checks if its a founder or not.

**Comment by alazymeme** [comment]
Can I add broadcaster support in here? Shouldn't need to be added to the refresh module. I just want to implement it for another pr :)
Basically to add the broadcaster badge OR the mod badge (depending on the scenario).

**Comment by pajlada** [comment]
Please revert the broadcaster column change

**Comment by alazymeme** [comment]
@pajlada All good now

File: pajbot/web/__init__.py [hunk 11]
```diff
@@ -143,7 +143,11 @@ def init(args):
             "deck_tab_images": config.getboolean("web", "deck_tab_images"),
             "websocket": {"host": config["websocket"].get("host", f"wss://{config['web']['domain']}/clrsocket")},
         },
-        "streamer": {"name": config["web"]["streamer_name"], "full_name": config["main"]["streamer"]},
+        "streamer": {
+            "name": config["web"]["streamer_name"],
+            "full_name": config["main"]["streamer"],
+            "id": StreamHelper.get_streamer_id,
+        },
```
RAnders00, line 150: this should not be in here

templates/user/username_link.html
```diff
@@ -1,7 +1,14 @@
-{% if user.moderator %}
+{% if user.id == streamer.id %}
+<div class="pui badge broadcaster"></div>
+{% elif user.moderator %}
```
RAnders00, line 3: this should not be in here

alazymeme: There is no record of the streamer being in the user models file. How would the bot import `user.streamer` if it isn't defined?

RAnders00: I thought it was said on stream that the broadcaster badge would be added separately. I want to merge this too, adding change after change after change infinitely delays the PR. Please remove the two marked changes and then we can get this PR merged. We can talk about this kind of stuff in a separate Pull request

alazymeme: No, pajlada said if it's badge-related then it can remain in this PR. For the stuff that's related to the https://github.com/pajbot/pajbot/pull/950 - it can be committed in that PR. The broadcaster badge on the userpage has nothing to do with the clip command pr.

RAnders00:
image
I just gave this PR a try on my dev setup, and analyzed the query. We will have to also create an index on the `vip` column, as you can see the query takes way too long, which is caused by the Seq Scan over the entire user table (as can be seen the by "actual time" the query analyzer outputs).

PR #887, opened 2020-05-21
Add option to unban when using the unpermaban command

### LINKED ISSUE #739: Unpermaban command should unban the person from chat

alazymeme:
Simple option that allows users to be unbanned from chat upon use of the `!unpermaban` command. Fixes #739 

Pull request checklist:


Felanbird:
I believe the unban should be followed by a 1 second timeout so an unban reason can be added for mod logs. (viewercard)

RAnders00:
The thing @Felanbird wrote can be added separately

alazymeme:
@RAnders00 Do you want me to 100% add it separately? Or just commit it now?

RAnders00:
You can add it now if you want @alazymeme I was just saying it doesn't *have* to be added to this PR

pajbot/modules/basic/permaban.py
```diff
@@ -78,6 +94,12 @@ def unpermaban_command(self, bot, source, message, **rest):
             if self.settings["unban_from_chat"] is True:
                 bot.unban(user)
 
+            if (
```
RAnders00: I think this is a bit unintutive but there has to be a short delay (2 or 3 seconds) between unban and timeout, otherwise the timeout will not work, you can try it out yourself in normal chat. You could just do "bot.execute_later"

pajbot/modules/basic/permaban.py
```diff
@@ -78,6 +94,12 @@ def unpermaban_command(self, bot, source, message, **rest):
...
+                self.settings["enable_send_timeout"] is True
+                and self.settings["unban_from_chat"] is True
+            ):
+                bot.timeout(user, 1, self.settings["timeout_reason"].format(source=source), once=True)
+
```
RAnders00:
```suggestion
            if self.settings["unban_from_chat"] is True:
                bot.unban(user)
                
                if self.settings["enable_send_timeout"] is True:
                    # This is delayed because you can't timeout a user immediately after
                    # unban (some caching effect on Twitch's end, the timeout errors as
                    # if the user was still banned if you don't add a delay)
                    # Delayed by 3 seconds here
                    bot.execute_delayed(
                        3,
                        bot.timeout,
                        user,
                        1,
                        self.settings["timeout_reason"].format(source=source),
                        once=True
                    )
```

RAnders00: Note I didn't check the format, it might need to be reformatted before it can be merged

pajbot/modules/basic/permaban.py
alazymeme: Are you sure? This worked perfectly when I tested it in my own chat.

image

RAnders00: Oh wow, I made an assumption there, it does look like that's been fixed on Twitch's end. A delay used to be necessary

pajbot/modules/basic/permaban.py
```diff
@@ -66,6 +91,12 @@ def unpermaban_command(bot, source, message, **rest):
 
             AdminLogManager.add_entry("Permaban remove", source, log_msg)
 
+            if self.settings["unban_from_chat"] is True:
+                bot.unban(user)
+
+            if self.settings["enable_send_timeout"] is True and self.settings["unban_from_chat"] is True:
+                bot.timeout(user, 1, self.settings["timeout_reason"].format(source=source), once=True)
+
```
RAnders00, line 99: However I'd still like these two blocks to be collapsed
```suggestion
            if self.settings["unban_from_chat"] is True:
                bot.unban(user)

                if self.settings["enable_send_timeout"] is True:
                    bot.timeout(user, 1, self.settings["timeout_reason"].format(source=source), once=True)

```

alazymeme: Wait you marked this as resolved. Did you want me to do it?

RAnders00: I didn‘t mean to, oops. Yes, if you apply this change I can approve this and merge it

alazymeme: I specifically added that extra statement - what is the point of the bot trying to time someone out for 1 second if they're still banned? Unless I'm missing what is actually being changed here (due to the horribly small code block).

RAnders00: The way you proposed it checks unban_from_chat twice. This is just a change to make the check for the unban_from_chat setting once and then put the check for enable_send_timeout one level deeper into that if block. 
Maybe you missed the fact that I added a level of indentation

RAnders00: A timeout still won‘t be done if the unban is disabled.

Issue #492, opened 2019-09-14
Make level needed to view/set specific module properties configurable


DatGuy1:

RAnders00:
First we need to solve https://github.com/pajbot/pajbot/issues/108

alazymeme:
@DatGuy1 What specific modules were you looking at? Since the initial issue that needed to be solved has been solved.

DatGuy1:
For example, moderation modules such as ASCII protection should probably require level 1000 while other more fun modules such as raffles and roulettes shouldn't

alazymeme:
As per this comment: https://github.com/pajbot/pajbot/pull/892#issuecomment-633221708
This won't be implemented despite my efforts.

RAnders00:
Due to pajbot [being moved into maintenance mode](https://github.com/pajbot/pajbot/blob/master/README.md) I'm closing this feature request.

PR #892, opened 2020-05-24
Add level requirements to certain modules

alazymeme:
Added level 1000 requirements to most moderation modules, upped some more critical ones to 1500.

Fixes https://github.com/pajbot/pajbot/issues/492

Pull request checklist:

### DIFF
--- a/pajbot/modules/pnsl.py
+++ b/pajbot/modules/pnsl.py
@@ -15,6 +15,7 @@ class PNSLModule(BaseModule):
     NAME = "Run P&SL lists"
     DESCRIPTION = "Run P&SL lists through the !runpnsl command"
     CATEGORY = "Moderation"
+    CONFIGURE_LEVEL = 1500
     SETTINGS = [
         ModuleSetting(
             key="level",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Minor: Changed level requirements for moderation and some admincommands modules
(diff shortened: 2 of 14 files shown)


alazymeme:
@DatGuy1 Are there anymore changes you wanted?

pajbot/modules/pnsl.py
```diff
@@ -15,6 +15,7 @@ class PNSLModule(BaseModule):
     NAME = "Run P&SL lists"
     DESCRIPTION = "Run P&SL lists through the !runpnsl command"
     CATEGORY = "Moderation"
+    CONFIGURE_LEVEL = 1500
```
Felanbird, line 18: I personally think this module should be 1000, people who utilize this module are most likely 1000 or 2000, and almost never 1500.

RAnders00:
Level 500 is "Moderator". 1000 is Broadcaster/Super Moderator I think.

So denying moderators access to moderation modules is wrong.

This change also breaks everyone's setup where they have their actual moderators as level 500.

I think a better solution would be to allow a different (lower) level to also have access to edit modules, but only the "fun" modules.

alazymeme:
> Level 500 is "Moderator". 1000 is Broadcaster/Super Moderator I think.
> 
> So denying moderators access to moderation modules is wrong.
> 
> This change also breaks everyone's setup where they have their actual moderators as level 500.
> 
> I think a better solution would be to allow a different (lower) level to also have access to edit modules, but only the "fun" modules.
> 
> I appreciate the effort though.

I completely disagree. This would be the same system as streamelements. Mods can enable/disable modules, but cannot modify unless they have direct access to the streamer's dashboard.

Issue #638, opened 2019-11-03
Add command to give every user points


RAnders00:
Original request form https://github.com/Parrallax (#622)

> Perhaps the ability to add points to all users at once.

alazymeme:
Does this include the ability to add points to individual users?

RAnders00:
> Does this include the ability to add points to individual users?

You can already do that with !editpoints and !setpoints

DatGuy1:
I have an existing implementation at https://github.com/admiralbullbot/bullbot/blob/master/pajbot/modules/viewerpoints.py

RAnders00:
Due to pajbot [being moved into maintenance mode](https://github.com/pajbot/pajbot/blob/master/README.md) I'm closing this feature request.

PR #921, opened 2020-06-20
Add a module to give points to all users currently in chat

### LINKED ISSUE #300: !awardpoints command
I think it would be good to have a command that just gives points intead of edit points,
Also have it so you could use !awardpoints \* 1000 to give everyone in chat 1k for example

alazymeme:
Made by @DatGuy1 & improved by @TroyDota (yeah I just yoinked the code and pr'ed it)
Fixes #300
Fixes https://github.com/pajbot/pajbot/issues/638

Pull request checklist:


pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+
+        userBasics = bot.twitch_helix_api.bulk_get_user_basics_by_login(currentChatters)
+
+        # filter out invalid/deleted/etc. users
+        userBasics = [e for e in userBasics if e is not None]
+
+        with DBManager.create_session_scope() as db_session:
+            subscribers = [
+                user.id
+                for user in db_session.query(User).filter_by(subscriber=True).all()
```
RAnders00, line 80: this right here is and will be a problem. loading all subscribers will lag the bot and place a big load on the database.

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+
+class MassPointsModule(BaseModule):
+    ID = __name__.split(".")[-1]
+    NAME = "Mass Points"
+    DESCRIPTION = "Give points to everyone watching the stream"
+    CATEGORY = "Feature"
+    SETTINGS = [
+        ModuleSetting(
+            key="sub_points",
+            label="Points to give for subscribers for each pleb point",
```
RAnders00, line 24: This label should be changed/clarified

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+            type="number",
+            required=True,
+            placeholder="1",
+            default=1,
+            constraints={"min_value": 0, "max_value": 100},
+        )
+    ]
+
+    def load_commands(self, **options):
+        self.commands["masspoints"] = Command.raw_command(
```
RAnders00, line 34: I don't necessarily agree with the command being `!masspoints`, it should be something more descript

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+            self.command_masspoints,
+            level=500,
+            description="Give a specific number of points to everyone watching the stream",
+            examples=[
+                CommandExample(
+                    None,
+                    "Give 300 points (for a fisting)",
+                    chat="user:!masspoints 300\n"
+                    "bot: user just gave 159 viewers 300 points! Enjoy FeelsGoodMan",
```
RAnders00: It shouldn't say `user` in the bot message, should probably say "pajlada" or something

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+                    description="Give 300 points to the number of people in chat, in this case 159",
```
RAnders00, line 43: Simplify this description or omit it altogether.

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+                ).parse()
+            ],
+        )
+
+    def command_masspoints(self, bot, source, message, **rest):
+        if not message:
+            return False
+
+        pointsArgument = message.split()[0]
```
RAnders00, line 52: should be explicit that we are splitting on space.

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+        givePoints = 0
```
RAnders00, line 53: no need to initialize it to 0 here

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+
+        try:
+            givePoints = int(pointsArgument)
+        except ValueError:
+            bot.whisper(source, "Invalid Usage, please specify a whole number.")
+            return False
+
+        # If they enter 0, there is no point in doing an update.
+        if givePoints == 0:
```
RAnders00, line 62: fix: should be a check for <=

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+            return
+
+        currentChatters = bot.twitch_tmi_api.get_chatter_logins_by_login(bot.streamer)
```
RAnders00, line 65: right now this is doing API requests and heavy database operations on the main event loop. This should be rewritten to make use of bot.action_queue.submit

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+        numUsers = len(currentChatters)
+        if not currentChatters:
+            bot.say("Error fetching chatters")
+            return False
```
RAnders00, line 69: What is this checking for? Empty list? I don't like these `if not <variable>:` style checks because it's extremely ambiguous what you are checking. Make this check explicit, e.g. `if len(variable) == 0:`

pajbot/modules/basic/masspoints.py
RAnders00, line 52: pointsArgument should be snake_case

pajbot/modules/basic/masspoints.py
RAnders00, line 65: currentChatters should be snake_case

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+        except ValueError:
+            bot.whisper(source, "Invalid Usage, please specify a whole number.")
+            return False
+
+        # If they enter 0, there is no point in doing an update.
+        if givePoints == 0:
+            return
+
+        currentChatters = bot.twitch_tmi_api.get_chatter_logins_by_login(bot.streamer)
+        numUsers = len(currentChatters)
```
RAnders00, line 66: numUsers should be snake_case

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+        if not currentChatters:
+            bot.say("Error fetching chatters")
+            return False
+
+        userBasics = bot.twitch_helix_api.bulk_get_user_basics_by_login(currentChatters)
```
RAnders00, line 71: userBasics should be snake_case

pajbot/modules/basic/masspoints.py
```diff
@@ -0,0 +1,105 @@
...
+ON CONFLICT (id) DO UPDATE SET
+    points = "user".points + :add_points
+WHERE "user".ignored = False and "user".banned = False and "user".num_lines > 5
+            """
+                ),
+                update_values,
+            )
+        bot.say(
+            f"{source} just gave {numUsers} viewers {givePoints} points each! Enjoy FeelsGoodMan"
+        )
```
RAnders00: If you rewrite this to use the action queue, remember to then do the `bot.say` calls via `bot.execute_now` to avoid multithreading problems, e.g. `bot.execute_now(bot.say, "abc def"))` instead of `bot.say("abc def")`

pajbot/modules/masspoints.py
```diff
@@ -21,7 +20,7 @@ class MassPointsModule(BaseModule):
     SETTINGS = [
         ModuleSetting(
             key="sub_points",
-            label="Points to give for subscribers for each pleb point",
+            label="Award this amount of points to subs for every pleb point",
```
RAnders00, line 23: Just make it "multiply points given to subscribers by this factor", sorry for not directly suggesting what I had in mind earlier

RAnders00: Additionally, now I realize this, but I think it would be nice to have this as a "percentage" setting where a setting of 100 gives plebs and subs the same amount. E.g. 200 would give subs twice the amount of points, etc. That way you can set it to something other than whole number multipliers.

pajbot/modules/basic/masspoints.py
TroyDota: Is a solution to select only their ids?

TroyDota:
> Additional changes
> 
> * the module should not be under the "basic" module/directory
> * Fetching the current chatters seems overkill for this module. Consider using the `last_seen` value in the database

Also this, wouldnt this also subsequently put a big strain on the database engine?

RAnders00:
Your intuition would be that scanning all users for a certain range of `last_seen` would be slow. However, if you put an index on `last_seen` it can be filtered extremely quickly, even for ranges.

This is also the reason I suggest to put big updates like this entirely in the database, because then you can run the whole thing through `EXPLAIN ANALYZE` to find bottlenecks. What you are thinking of would show up as a `Seq Scan`, a sequential scan where each record is considered one-by-one and checked.

I'm fairly sure we might already have an index on `last_seen` due to other usages such as the mass ping module. If the index doesnt exist yet, it can be added with a database migration.

Issue #911, opened 2020-06-11
Alert when the streamer goes live


alazymeme:
People like this on streamelements for some reason, not a bad idea to add it to pajbot
![image](https://user-images.githubusercontent.com/12804673/84383519-c5313c00-ac2f-11ea-8f15-97501244bc86.png)

Issue #700, opened 2019-11-20
Alerts module


alazymeme:
Merge sub alerts and add more options to a broader alerts module

- [x] Raid alerts
- [x] New chatter alerts
- [x] Live alerts
- [ ] Streamlabs donation alerts
- [ ] Streamelements donation alerts
- [ ] Host alerts
- [ ] Bit cheering alerts
- [ ] Follow alerts
- [ ] Channel points alert

RAnders00:
Due to pajbot [being moved into maintenance mode](https://github.com/pajbot/pajbot/blob/master/README.md) I'm closing this feature request. However you are of course free to submit pull requests to add these alert modules in the future.

PR #924, opened 2020-06-22
Added a live alerts module & capitalized streamer name variable

alazymeme:
Pull request checklist:

image

Chat can now be notified when their favourite streamer goes live by having pajbot post a message in chat. Also added a variable that pulls the capitalized streamer's name from the config - I needed this change in order to make the module look nice :)

Closes https://github.com/pajbot/pajbot/issues/911
Related: https://github.com/pajbot/pajbot/issues/700


alazymeme:
> No objections! This looks very nice! Thank you @alazymeme 👍
> I will leave it open though so @pajlada can review it before I merge this

Just thought about it. Should the variable be consistent with `$(tb:broadcaster)` or can I rename the broadcaster variable to `$(tb:channel)` or `$(tb:streamer)` instead? 🤔

RAnders00:
It could be `$(broadcaster:name)`, `$(broadcaster:id)`, `$(broadcaster:login)` in the future but in my opinion this can be fixed separately, also you have to remember to keep the $(tb:broadcaster) variable alive to not break existing commands

Issue #928, opened 2020-07-04
no-referrer-when-downgrade when trying to visit the bot_login url while setting up the bot


PrestonRoberts:
I followed the detailed installation docs and when i went to https://streamer_name.your-domain.com/bot_login where the streamer_name and domain name are my own the web page doesn't load but the error it gives me is "no-referrer-when-downgrade". The sudo nginx -t passes and status says everything is good. The error logs for nginx also don't say anything. I used a domain from google and used the cloudflare tutorial. Sorry if I am being dumb and I just missed something but how can I troubleshoot this/where should I be looking to fix this issue?

PrestonRoberts:
After doing some research on other issues I was also able to find this error

pajbot-web@poberts[44413]: *** Operational MODE: single process ***
pajbot-web@poberts[44413]: Traceback (most recent call last):
pajbot-web@poberts[44413]:   File "./app.py", line 20, in <module>
pajbot-web@poberts[44413]:     pajbot.web.init(args)
pajbot-web@poberts[44413]:   File "./pajbot/web/__init__.py", line 75, in init
pajbot-web@poberts[44413]:     with open(args.config, "w") as configfile:
pajbot-web@poberts[44413]: PermissionError: [Errno 13] Permission denied: 'configs/poberts.ini'
pajbot-web@poberts[44413]: unable to load app 0 (mountpoint='') (callable not found or import error)
pajbot-web@poberts[44413]: *** no app loaded. GAME OVER ***
pajbot-web@poberts[44413]: VACUUM: unix socket /var/run/pajbot/poberts/web.sock removed.
pajbot-web@poberts.service: Main process exited, code=exited, status=22/n/a
pajbot-web@poberts.service: Failed with result 'exit-code'.

And also this 

● pajbot-web@poberts.service - pajbot-web for poberts
     Loaded: loaded (/etc/systemd/system/pajbot-web@.service; enabled; vendor preset: enabled)
     Active: activating (auto-restart) (Result: exit-code) since Sat 2020-07-04 02:41:08 EDT; 472ms ago
    Process: 44674 ExecStart=/opt/pajbot/venv/bin/uwsgi --ini uwsgi_shared.ini --socket /var/run/pajbot/poberts/web.sock --pyargv --config configs/poberts.ini --virtualenv venv (code=exited, status=22)
   Main PID: 44674 (code=exited, status=22)

pajbot-web@poberts.service: Main process exited, code=exited, status=22/n/a
[...]

PR #929, opened 2020-07-04
Fix retry logic if getting the login token fails

RAnders00:
I tested that it works. It just spams the NoTokenError every 2 seconds until you complete the bot_login process.

Pull request checklist:


RAnders00:
Note that I saw it in the logs of #928 but it doesn't fix that issue, it's unrelated to his issue but should be fixed nonetheless.

PR #932, opened 2020-07-10
Miscellaneous fixes

alazymeme:
Pull request checklist:

- Fixed emotes that are no longer emotes
- Paid timeout only times out once
- Added strings for raffle emotes and "points" grammar


alazymeme:
@RAnders00 or @pajlada Could you please help me on this? I'm struggling to figure out the correct syntax and location.
I want dynamic emotes in the raffle. So if a user loses points, it uses LUL. If a user gains points, it uses PogChamp.
Similarly for the points formatting. if 1 point has been lost or won, then it should be formatted as `point`, not `points`.

PR #933, opened 2020-07-11
Add read/edit level requirements to modules

alazymeme:
Pull request checklist:


alazymeme:
I've configured the cosmetic changes, but the `hidden` attribute can easily be bypassed using the devtools in your browser (simply removing the attribute, changing the value and saving with edit it as per normal)

templates/admin/configure_module.html
```diff
@@ -52,20 +52,59 @@ <h3>Settings</h3>
     <div class="{{ 'required' if setting.required else ''}} {{ 'inline' if setting.type in ['boolean'] else '' }} field eight wide">
         <label for="key_{{ setting.key }}">{{ setting.label }}</label>
         {% if setting.type == 'text' %}
-        <input type="text" " id="key_{{ setting.key }}" name="{{ setting.key }}" placeholder="{{ setting.placeholder }}" value="{{ module.settings[setting.key] }}" />
+            {% if session.user.level < setting.required_write_level %}
+        	    <input type="text" " id="key_{{ setting.key }}" name="{{ setting.key }}" placeholder="{{ setting.placeholder }}" value="{{ module.settings[setting.key] }}" disabled />
+            {% elif session.user.level < setting.required_read_level %}
+        	    <input type="text" " id="key_{{ setting.key }}" name="{{ setting.key }}" placeholder="{{ setting.placeholder }}" value="{{ module.settings[setting.key] }}" hidden />
```
pajlada, line 58: The value must not be set if the user is not high level enough to read the setting

alazymeme:
> I was thinking that maybe if the user doesn't have edit permission, the should just be a read-only equivalent so it's not sent with the form request at all

Just trying to figure out what this means. Could you please elaborate? If the key isn't sent with the form request, then how is it read-only? Would that not be the same as not having the view permission? Or am I missing something?

RAnders00:
I'm sorry to have to say this, but because pajbot was moved into maintenance mode we don't want to accept major internal changes anymore.

Issue #930, opened 2020-07-05
Add a percentage field to the case checker module


alazymeme:
So it can be used as an actual caps protection module, as opposed to only a troll module.

PR #941, opened 2020-07-21
Refactor the case-checker module

alazymeme:
Currently the case checker module acts as a "troll" or "joke" module. This change will give it a functional aspect to it.

Pull request checklist:

Fixes https://github.com/pajbot/pajbot/issues/930

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added a secondary optional message to send when the channel goes live (live alerts module) (#938)
 - Minor: Added a new module to print a chat alert when a user announces they are new (#926)
 - Minor: Added a live alerts module to notify chat when the streamer goes live (#924)
 - Minor: Added a `streamer_display` variable to show the capitalized version of the broadcaster (#924)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Minor: Added a percentage and max-based uppercase timeout feature. (#941)
+- Minor: Refactored the case checker module. It now allows for more custom settings for both lowercase and uppercase timeouts. (#941)
 - Minor: Added a secondary optional message to send when the channel goes live (live alerts module) (#938)
 - Minor: Added a new module to print a chat alert when a user announces they are new (#926)
 - Minor: Added a live alerts module to notify chat when the streamer goes live (#924)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added a percentage and max-based uppercase timeout feature. (#941)
 - Minor: Added a secondary optional message to send when the channel goes live (live alerts module) (#938)
 - Minor: Added a new module to print a chat alert when a user announces they are new (#926)
 - Minor: Added a live alerts module to notify chat when the streamer goes live (#924)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Minor: Refactored the case checker module. It now allows for more custom settings for both lowercase and uppercase timeouts. (#941)
+- Minor: Refactored the case checker module - this changes the default behaviour of the module by forcing usage of a max/percentage-based upper/lowercase setting as opposed to timing out ANY upper/lowercase characters in a sentence. If the max upper/lowercase setting is met, the user is immediately timed out. If not, the bot will check the total upper/lowercase characters in a sentence to ensure it is above the minimum; if so, it will timeout based on if the percentage of upper/lowercase characters is above the pre-defined percentage. Users are able to re-create the original behaviour by simply setting the max upper/lowercase setting to 0. (#941)
 - Minor: Added a secondary optional message to send when the channel goes live (live alerts module) (#938)
 - Minor: Added a new module to print a chat alert when a user announces they are new (#926)
 - Minor: Added a live alerts module to notify chat when the streamer goes live (#924)
--- a/pajbot/modules/casechecker.py
+++ b/pajbot/modules/casechecker.py
@@ -53,7 +53,7 @@ class CaseCheckerModule(BaseModule):
             type="text",
             required=False,
             placeholder="",
-            default="Too many lowercase characters ",
+            default="Too many lowercase characters",
             constraints={},
         ),
         ModuleSetting(
@@ -67,7 +67,7 @@ class CaseCheckerModule(BaseModule):
         ),
         ModuleSetting(
             key="lowercase_percentage",
-            label="Maximum percent of uppercase letters allowed in message",
+            label="Maximum percent of lowercase letters allowed in message",
             type="number",
             required=True,
             placeholder="",
--- a/pajbot/modules/casechecker.py
+++ b/pajbot/modules/casechecker.py
@@ -25,7 +25,7 @@ class CaseCheckerModule(BaseModule):
             type="boolean",
             required=True,
             default=False,
-        ), 
+        ),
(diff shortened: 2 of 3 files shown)


pajbot/modules/casechecker.py
```diff
@@ -86,6 +129,23 @@ def on_message(self, source, message, **rest):
...
             return False
+        # The module will check first if the amount of uppercase letters in a message is more than the max set amount
+        # If not, the module will figure out the percentage of uppercase letters in the message
+        # If the percentage is higher than the max percent, then the user will be timed out
+        amount_capitals = sum(1 for c in message if c.isupper())
+        if self.settings["timeout_percentage_toggle"] is True:
+            if amount_capitals >= self.settings["max_amount"]:
+                self.bot.timeout(
+                    source, self.settings["timeout_duration"], reason=self.settings["percentage_timeout_reason"]
+                )
```
RAnders00, line 140: forgot to `return False` here

alazymeme:
> the `timeout_uppercase` and `max_amount` = 0 do the same thing.
> 
> Please consider instead making the percentage and absolute-based settings available for both uppercase and lowercase, and then removing the old settings
> 
> Basically, replace `timeout_uppercase` and `timeout_lowercase` with respective max. percentage and absolute amounts.
> 
> Also, the current settings names you chose are too vague, as well as the description. It should be more clear that those settings are for **uppercase** letters. What kind of setting description is "Maximum amount allowed in a message"?

Your comment about the essentially duplicated code made me want to make the entire thing a lot more custom. So I've separated upper/lowercase and added a few more options to each. I'm looking to add vip support once https://github.com/pajbot/pajbot/pull/886 is done

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
             constraints={"min_value": 3, "max_value": 120},
         ),
         ModuleSetting(
-            key="online_chat_only", label="Only enabled in online chat", type="boolean", required=True, default=True
+            key="lowercase_timeout_reason",
+            label="Lowercase Timeout Reason",
+            type="text",
+            required=False,
+            placeholder="",
+            default="Too many lowercase characters ",
```
RAnders00, line 56: you accidentally put a trailing space here
```suggestion
            default="Too many lowercase characters",
```

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
+            constraints={},
```
RAnders00, line 57: should probably consider putting some reasonable maximum length restriction on this

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
+        ),
+        ModuleSetting(
+            key="min_lowercase_characters",
+            label="Minimum amount of lowercase characters before checking for a percentage",
+            type="number",
+            required=True,
+            placeholder="",
+            default=8,
+            constraints={},
```
RAnders00, line 66: I think you want constraints here too, otherwise you can enter negative numbers and numbers larger than 500

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
+            constraints={"min_value": 0, "max_value": 100},
+        ),
+        ModuleSetting(
+            key="max_lowercase",
+            label="Maximum amount of lowercase characters allowed in a message",
+            type="number",
+            required=True,
+            placeholder="",
+            default=50,
+            constraints={},
```
RAnders00, line 84: same thing here, the setting should be between 0 and 500

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
+            label="Minimum amount of lowercase characters before checking for a percentage",
+            type="number",
+            required=True,
+            placeholder="",
+            default=8,
+            constraints={},
+        ),
+        ModuleSetting(
+            key="lowercase_percentage",
+            label="Maximum percent of uppercase letters allowed in message",
```
RAnders00, line 70:
```suggestion
            label="Maximum percent of lowercase letters allowed in message",
```
copy paste error?

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
-            type="text",
-            required=False,
+            key="min_uppercase_characters",
+            label="Minimum amount of uppercase characters before checking for a percentage",
+            type="number",
+            required=True,
             placeholder="",
-            default="NO LOWERCASE CHARACTERS ALLOWED",
+            default=8,
+            constraints={},
```
RAnders00, line 114: same thing here, should be constrained to be between 0 and 500

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
+            constraints={"min_value": 0, "max_value": 100},
+        ),
+        ModuleSetting(
+            key="max_uppercase",
+            label="Maximum amount of uppercase characters allowed in a message",
+            type="number",
+            required=True,
+            placeholder="",
+            default=50,
             constraints={},
```
RAnders00, line 132: should also add constraint here

pajbot/modules/casechecker.py
```diff
@@ -37,33 +36,99 @@ class CaseCheckerModule(BaseModule):
...
         ),
         ModuleSetting(
             key="uppercase_timeout_reason",
             label="Uppercase Timeout Reason",
             type="text",
             required=False,
             placeholder="",
-            default="no uppercase characters allowed",
+            default="Too many uppercase characters",
             constraints={},
```
RAnders00, line 105: a constraint should be added here too

RAnders00:
circleci broken again. Will have to wait for @pajlada to fix it or temporarily bypass the CI restriction

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Refactored the case checker module. It now allows for more custom settings for both lowercase and uppercase timeouts. (#941)
```
pajlada, line 5: Please define what new functionality has been added to the module

pajbot/modules/casechecker.py
```diff
@@ -37,55 +37,160 @@ class CaseCheckerModule(BaseModule):
...
+            default="Too many lowercase characters",
+            constraints={"max_str_len": 500},
+        ),
+        ModuleSetting(
+            key="min_lowercase_characters",
+            label="Minimum amount of lowercase characters before checking for a percentage",
+            type="number",
+            required=True,
+            placeholder="",
+            default=8,
```
pajlada, line 66: This changes the default behaviour of the module - we need to either:
a) Not change the default behaviour of the module, or
b) Clearly define in the Changelog that we have changed the default behaviour of the module.
If module settings are checked, but the module is just enabled and it doesn't behave as expected that's not a fun time.

pajbot/modules/casechecker.py
```diff
@@ -37,55 +37,160 @@ class CaseCheckerModule(BaseModule):
...
+        ),
+        ModuleSetting(
+            key="max_uppercase",
+            label="Maximum amount of uppercase characters allowed in a message",
+            type="number",
+            required=True,
+            placeholder="",
+            default=50,
+            constraints={"min_value": 0, "max_value": 500},
         ),
```
pajlada, line 134: These settings should change order to fit with the functionality better.
Right now, the `max_uppercase` and `max_lowercase` settings are the ones that take priority, yet they're at the bottom of the configuration.
It might also be worth adding to the `max_uppercase/lowercase` descriptions that they are the values that are checked first

PR #947, opened 2020-07-26
Fix level display on bttv/ffzemotes detailed commands page

alazymeme:
Pull request checklist:


RAnders00:
this doesn't only fix the web interface. the settings just weren't used at all before. maybe that should be reflected in the changelog entry

PR #950, opened 2020-07-28
Create a clip command module

alazymeme:
Pull request checklist:

I created a clip command module that simply allows the viewer to type `!clip` and it will clip the last 30 seconds of the stream.


alazymeme:
The problem I'm having at the moment, is that the clip is created, but twitch inherently is returning nothing when fetching the clip url:
image

Could anyone give some insight as to why this would be happening and how I would go about fixing it?

alazymeme:
PR is dependent on https://github.com/pajbot/pajbot/pull/886

RAnders00:
Mind if I mark this as a draft in the meantime?

alazymeme:
> Mind if I mark this as a draft in the meantime?

No

alazymeme:
Decided not to implement discord webhook functionality yet, will work on it at a later date.

Issue #919, opened 2020-06-16
Webpage has levels labelled inconsistently


alazymeme:
I used some mixed examples of nymn and my own webpage:

Nymn's page:
![image](https://user-images.githubusercontent.com/12804673/84770034-3864f480-b01a-11ea-8134-05075b6c26a1.png)

![image](https://user-images.githubusercontent.com/12804673/84770105-55012c80-b01a-11ea-9bc7-8e7b1c63861d.png)

My own page:
According to the moderators page, 1500 should be labelled as Super Mod/Broadcaster; not Admin:
![image](https://user-images.githubusercontent.com/12804673/84770153-6813fc80-b01a-11ea-9ad0-22ec9f88d9ab.png)

PR #968, opened 2020-08-11
Separate super mods from broadcasters on the user page

alazymeme:
Pull request checklist:

Broadcasters should appear to have more perms than a super mod, but less than an admin (since an admin has eval perms)

Fixes https://github.com/pajbot/pajbot/issues/919


alazymeme:
@pajlada

RAnders00:
is it possible to condense your two pull requests into one, and then you just make the changelog entry "Minor: Made names for user levels on the website consistent"

PR #972, opened 2020-08-13
Add a dependabot file

alazymeme:
Pull request checklist:

Allows the creation of PRs in order to keep dependencies updated. It will check `pyproject.toml`, `requirements-dev.txt` and `requirements.txt`. I looked into ignoring the pyproject file, but I can't see it being possible - not a big issue since there's no dependencies in there anyway.

Docs can be found here: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates


RAnders00:
Good idea, thanks for taking care of it!

PR #978, opened 2020-08-16
Disallow the pnslrun command from being used while the streamer is live

alazymeme:
Pull request checklist:

PNSL lists should not be run while the streamer is online. This will aid for users who opt to use the controlling channel for running these lists.

### DIFF
--- a/pajbot/modules/pnsl.py
+++ b/pajbot/modules/pnsl.py
@@ -62,10 +62,9 @@ class PNSLModule(BaseModule):
                 self.pnsl_token = bot.config["pnsl"].get("token", None)
 
     def run_pnsl(self, bot, source, message, event, args):
-        if self.settings["offline_only"] is True:
-            if not self.bot.is_online:
-                bot.whisper(source, f"{bot.streamer} is online! Skipping PNSL list eval.")
-                return False
+        if self.settings["offline_only"] and self.bot.is_online:
+            bot.whisper(source, f"{bot.streamer_display} is live! Skipping PNSL list eval.")
+            return False
 
         base_url = "https://bot.tetyys.com/api/v1/BotLists"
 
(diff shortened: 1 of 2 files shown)


Felanbird:
This should be an option, the only reason this would be needed is if you had some sort of plan to auto-run the lists via !runpnsl. I can think of instances where running a list while live would be needed, and at the speed runpnsl executes, it wouldn't interrupt chat that much anyway.

pajbot/modules/pnsl.py
```diff
@@ -55,6 +62,11 @@ def __init__(self, bot):
                 self.pnsl_token = bot.config["pnsl"].get("token", None)
 
     def run_pnsl(self, bot, source, message, event, args):
+        if self.settings["offline_only"] is True:
+            if not self.bot.is_online:
+                bot.whisper(source, f"{bot.streamer} is online! Skipping PNSL list eval.")
+                return False
```
RAnders00, line 68:
```suggestion
        if self.settings["offline_only"] and self.bot.is_online:
            bot.whisper(source, f"{bot.streamer_display} is live! Skipping PNSL list eval.")
            return False
```

First, `not self.bot.is_online` would have exactly the opposite effect (disallowing the run when offline), second you can collapse the nested if-blocks into one. Also, we should use the streamer's display name.

pajbot/modules/pnsl.py
```diff
@@ -16,6 +16,13 @@ class PNSLModule(BaseModule):
     DESCRIPTION = "Run P&SL lists through the !runpnsl command"
     CATEGORY = "Moderation"
     SETTINGS = [
+        ModuleSetting(
+            key="offline_only",
+            label="Only allow the PNSL commands to be run while the stream is offline",
+            type="boolean",
+            required=True,
+            default=False,
+        ),
```
RAnders00, line 25: should move the setting down in the list of settings, below `chunk_delay` at the end.

Issue #967, opened 2020-08-11
Uncaught exception while refreshing vips


Leppunen:
I'm encountering the following exception after the bot has been running for a few minutes.

The channel has no vips available and the bot is running on it's own channel for development purposes.

Python version 3.8.5 running in Alpine linux inside docker.

```python
[2020-08-11 10:23:16,393] [ERROR  ] Logging an uncaught exception (ActionQueue)
Traceback (most recent call last):
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1179, in _execute_context
    context = constructor(dialect, self, conn, *args)
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 681, in _init_compiled
    self.compiled_parameters = [
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 682, in <listcomp>
    compiled.construct_params(m, _group_number=grp)
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/sql/compiler.py", line 694, in construct_params
    raise exc.InvalidRequestError(
sqlalchemy.exc.InvalidRequestError: A value is required for bind parameter 'id' (Background on this error at: http://sqlalche.me/e/cd3x)
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/app/pajbot/utils/time_method.py", line 18, in wrap
    ret = f(*args, **kwargs)
  File "/app/pajbot/modules/vip_refresh.py", line 84, in _process_vip_logins
    db_session.execute(
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 1268, in execute
    return self._connection_for_bind(bind, close_with_result=True).execute(
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 988, in execute
    return meth(self, multiparams, params)
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/sql/elements.py", line 287, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/app/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1101, in _execute_clauseelement
    ret = self._execute_context(
[...]

PR #981, opened 2020-08-16
Fix vip refresh in case a channel has zero VIPs

pajlada:
Pull request checklist:


pajlada:
Fixes #967

pajlada:
No changelog entry since no change has been made since the last release

PR #989, opened 2020-08-17
Fix new P&SL list parsing

### LINKED ISSUE #988: Incorrectly Parse UTF8 for the purposes of runpnsl
image

```
/app/venv/lib/python3.8/site-packages/sqlalchemy/orm/query.py:196: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if entities is not ():
[2020-08-17 11:33:24,393] [ERROR  ] Logging an uncaught exception (IRC event handler)
Traceback (most recent call last):
  File "/app/pajbot/managers/irc.py", line 149, in _dispatcher
    method(conn, event)
  File "/app/pajbot/bot.py", line 799, in on_pubmsg
    self.parse_message(event.arguments[0], source, event, tags=tags)
  File "/app/pajbot/bot.py", line 720, in parse_message
    command.run(self, source, remaining_message, event=event, args=extra_args, whisper=whisper)
  File "/app/pajbot/models/command.py", line 417, in run
    self.run_action(bot, source, message, event, args)
  File "/app/pajbot/models/command.py", line 424, in run_action
    ret = self.action.run(bot, source, message, event, args)
  File "/app/pajbot/models/action.py", line 247, in run
    return self.cb(bot=bot, source=source, message=message, event=event, args=args)
  File "/app/pajbot/modules/pnsl.py", line 82, in run_pnsl
    error_data = res.json()
  File "/app/venv/lib/python3.8/site-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/local/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.8/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

pajlada:
When fetching a PNSL list, split on the Line Feed character (\n) instead of using the splitlines helper.

Fixes #988 

Pull request checklist:

### DIFF
--- a/pajbot/modules/pnsl.py
+++ b/pajbot/modules/pnsl.py
@@ -83,7 +83,7 @@ class PNSLModule(BaseModule):
             bot.whisper(source, f"Something went wrong with the P&SL request: {error_data['errors']['Guid'][0]}")
             return False
 
-        privmsg_list = res.text.splitlines()
+        privmsg_list = res.text.split("\n")
 
         log.info(f"[P&SL] User {source.name} running list {guid} with {len(privmsg_list)} entries")
 
--- a/pajbot/modules/pnsl.py
+++ b/pajbot/modules/pnsl.py
@@ -73,6 +73,7 @@ class PNSLModule(BaseModule):
             return False
 
         guid = message.replace("https://bot.tetyys.com/BotList/", "")
+        guid = message.replace("https://bot.tetyys.com/api/v1/BotList/", "")
 
         headers = {"Authorization": f"Bearer {self.pnsl_token}"}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Bugfix: Handle new format of P&SL lists (#988, #989)
+
 ## v1.47
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!


pajbot/modules/pnsl.py
```diff
@@ -73,6 +73,7 @@ def run_pnsl(self, bot, source, message, event, args):
             return False
 
         guid = message.replace("https://bot.tetyys.com/BotList/", "")
```
RAnders00, line 75: huh? shouldnt this line be removed

5E7EN: If anything, this line should stay but the one under it (76) should go/be corrected. 
Being an /api link means it will be formatted like so: `https://bot.tetyys.com/api/v1/BotList/{guid}?Token={token}` - therefore not properly trimming the `Token` param. Unless I'm missing something here?

alazymeme: image
image
image

RAnders00: I mean like there's two assignments to `guid=`, the first one will always be overridden by the second one, no?

5E7EN: Valid point in that case. I thought something like this wouldn't matter in Python (which I have little experience with), but I stand corrected.

alazymeme: My understanding was that it works like a variable in javascript

PR #979, opened 2020-08-16
Bump all dependencies (where possible)

alazymeme:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable
- [x] Documentation in docs/ or install-docs/ was updated, if applicable

While dependabot is nice, randers mentioned concerns of dependabot opening a pr for each dependency. I've bumped all dependencies in this pr. Will work on testing. Perhaps @pajlada can start testing prior to release?

PR #990, opened 2020-08-17
fix leaguerank module for new riotwatcher version

RAnders00:
effort towards #979 

I honestly believe the leaguerank was completely broken before. It couldn't have possibly worked like this (the riotwatcher version in requirements was way too new for the code)

Pull request checklist:


RAnders00:
i couldnt be bothered to test it beyond the most basic usage, if there is still demand for this module today then more testing should be done

pajbot/modules/leaguerank.py
```diff
@@ -129,44 +142,37 @@ def league_rank(self, bot, source, message, **rest):
...
             bot.say(
-                f"The Summoner {summoner_name} on region {region.upper()} is currently in {tier} {division} with {league_points} LP 4Head"
+                f"The Summoner {summoner_name} on region {region.upper()} is currently in {tier} {rank} with {league_points} LP 4Head"
             )
-        except LoLException as e:
-            if e == error_429:
+        except ApiError as e:
+            log.exception("babyrage")
+            if e.response.status_code == 429:
                 bot.say(f"Too many requests. Try again in {e.headers['Retry-After']} seconds")
```
RAnders00, line 173: now that I see it, this will have to probably be e.response.headers

alazymeme: image

PR #1001, opened 2020-08-29
Migrate Game/Title fetching from Kraken v5 to Helix

### LINKED ISSUE #998: Move game + title fetching to helix
The current implementation isn't the best. When I use a game/title variable, I want to fetch the game/title at that current time. We shouldn't be requiring 2 separate variables.

pajlada:
This PR outgrew itself a bit, instead of only doing Game/Title fetching in Helix instead of Kraken v5, we also fetch video/stream information from Helix, and we use Helix for setting Game/Title.

During development, I started using mypy for getting early warnings in case I was sending the wrong parameters in somewhere, and it spreads very easily so there's a commit that adds typing to a few things that aren't related to this migration. If needed, this can be made into its own PR (including the addition of mypy to `requirements-dev.txt`)

Pull request checklist:

Fixes #998

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,17 @@
 
 ## Unversioned
 
+Because the Game/Title setting API calls are now using the Helix calls, it's no longer possible to use the Bot token to update the game/title of a channel, instead the Streamer token **must** be used. In addition to this, the Streamer token needs a new permission `user:edit:broadcast`.  
+In short: The Streamers must re-authenticate with the `/streamer_login` endpoint for `!setgame` and `!settitle` to work.
+
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Breaking: Replace Kraken Game/Title setting API calls with Helix ones. (#1001)
 - Minor: Added option to select your referred gambling command. (#1067)
 - Minor: The `sub_only` command option will now show in the `!debug command` command. (#1027)
 - Minor: The permaban module will now ban immediately on command use (#1014)
 - Minor: Added settings to change command name and cooldowns for showemote module. (#1007)
+- Minor: Replace Kraken Game/Title/Stream/Video fetching API calls with Helix ones (#1001)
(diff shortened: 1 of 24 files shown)


alazymeme:
Any news on this PR? peepoZoom

alazymeme:
@RAnders00 can you review pls alazymeme

RAnders00 (approved):

pajlada:
Will merge next weekend while letting it run on my bots - will also think about the office() comment

RAnders00 (changes requested):
This will need to be updated, right? https://github.com/pajbot/pajbot/blob/ce23381abd10fb3208d3c1b0a9c4b6b1bc45b163/templates/home.html#L8 Since the "online" key was removed.

pajlada:
> This will need to be updated, right?
> 
> https://github.com/pajbot/pajbot/blob/ce23381abd10fb3208d3c1b0a9c4b6b1bc45b163/templates/home.html#L8
> Since the "online" key was removed.

No I don't think so, the removed online key from the `UserStream` class is unused as far as I can tell. This `stream_data` object is fetched from Redis from here https://github.com/pajbot/pajbot/blob/fix/migrate-game-title-fetching-from-kraken-to-helix/pajbot/web/routes/base/home.py#L20-L23 and populated from https://github.com/pajbot/pajbot/blob/fix/migrate-game-title-fetching-from-kraken-to-helix/pajbot/models/stream.py#L283

pajlada:
After testing, the only thing I noticed was that the bot needs a restart when the Streamer has reauthenticated through the `/streamer_login` route to provide the required scope for `!settitle` and `!setgame` (or you could run `!eval bot.streamer_access_token_manager.initialize()` (maybe `.refresh()` is enough). I don't feel like that fits within the scope of this PR though

alazymeme:
Merge soon? PauseChamp

RAnders00 (approved):
👍 Merge it when you think it's been tested for long enough FarmingCommits

alazymeme:
PauseChamp

PR #1037, opened 2020-10-04
Disable mod_only bypass if command requires a level >= 500

### LINKED ISSUE #1035: Setting both mod_only and level parameters prefers one over the other
Set a command as mod_only with a level requirement of 500:

- A mod that is level 100 will not be able to use the command
- A non-mod that is level 500 is able to use the command

It should be one requirement or the other, not level taking priority over mod.

RAnders00:
Fixes #1035

Pull request checklist:


alazymeme:
The best way to test this is to add a new command with modonly and level 500, have the user be a mod but not level 500. The command still won't work ALazyMeme

RAnders00:
If a command is set to level 500 and mod_only then I would expect only twitch moderators with level 500 to be able to run the command. Why would it be an "OR" between the two instead of an "AND"?

pajlada:
wontfix

PR #1057, opened 2020-10-28
Update to major dependency versions too

RAnders00:
Pull request checklist:


alazymeme:
There's no reason to add this for pajbot1. GitHub actions doesn't support the `versioning-strategy` parameter and pip uses either `lockfile-only` or `auto`, both of which are useless (since `lockfile-only` restricts further and `auto` is the default regardless): https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#versioning-strategy

PR #1088, opened 2020-12-02
Make commands doc

alazymeme:
Pull request checklist:


RAnders00 (changes requested):
Not the right place to add this. The variables.txt was only to document how substitution works. Could you make a separate document about commands?

docs/commands.md
```diff
@@ -0,0 +1,13 @@
+# Command Arguments
+
+## Banphrases
+
+`--length`/`--time`/`--duration` - sets the timeout length of for the phrase (in seconds). Default = 600 seconds\
```
RAnders00, line 5: I would prefer if you could either use a bullet point list or use the "double space at end of line" instead of the backslash to create the linebreaks.

RAnders00 (changes requested):

docs/commands.md
alazymeme: Using a bullet point list wouldn't be consistent with the variables doc. I've used the double spaces.

RAnders00 (approved):

PR #1093, opened 2020-12-08
Banphrase module changes

alazymeme:
Pull request checklist:

I don't think there's a reason for the banphrase module to be displayed in the modules list. If you don't want banphrases, then don't add them; there's no reason to have the option to enable/disable the module.


RAnders00 (approved):
I generally agree, but will wait on @pajlada's thoughts too

pajlada:
Not worth the potential breaking change of running bots imo

PR #1122, opened 2021-01-10
Fix double message handling in safe_privmsg and is_bad_message functions

alazymeme:
Pull request checklist:

Initially the `safe_privmsg` function had it's own banphrase checking in it. This isn't required as there is a specific function for checking for bad messages: `is_bad_message`. This pr removes the double banphrase handling and makes use of the function that is already in place. I've also changed the behaviour of the `safe_privmsg` function; it will now not respond to command usage if it triggers a banphrase - which replicates the behaviour of the `safe_me` function (making the bot more consistent).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Minor: Fixed double banphrase handling in the `safe_privmsg` function and `is_bad_message` function.
+- Minor: Fixed double banphrase handling in the `safe_privmsg` function and `is_bad_message` function. (#1122)
 
 ## v1.48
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Fixed double banphrase handling in the `safe_privmsg` function and `is_bad_message` function. (#1122)
```
RAnders00, line 5: I don't think this actually needs a changelog entry because it doesn't change any behaviour.

RAnders00 (changes requested):

RAnders00:
Other than that, good PR!

RAnders00 (approved):

Issue #1118, opened 2021-01-02
run_through_banphrases parser doesn't work for modules


alazymeme:
Have tried adding it to both the math module and the lastfm module. The bot still sends the response as if it had never checked for the banphrase.

Relevant code:
https://github.com/alazymeme/pajbot/blob/8a0622c62da522bdbad7fca1180a4c2a29d46940/pajbot/modules/math.py#L86-L94

RAnders00:
I don‘t think that‘s what the flag is used for. In modules you‘d use bot.safe_privmsg and bot.safe_me, e.g.: https://github.com/pajbot/pajbot/blob/c3058eb1b71468b180871042eccebbe144fc845e/pajbot/bot.py#L628

if needed you could also add a safe_say method which is like say (escapes chat commands like `/timeout` from accidentally being executed)

RAnders00:
Example from trivia module: https://github.com/pajbot/pajbot/blob/c3058eb1b71468b180871042eccebbe144fc845e/pajbot/modules/trivia.py#L162

alazymeme:
Cheers, looks like it works

PR #1123, opened 2021-01-10
Add safe say function

alazymeme:
Pull request checklist:

Similar to the `safe_privmsg` and `safe_me` functions, the `safe_say` function will check the response against the banphrases before posting the response. If it is a banned phrase, it won't post anything at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Added a `safe_say` function.
+
 ## v1.48
 
 Because the Game/Title setting API calls are now using the Helix calls, it's no longer possible to use the Bot token to update the game/title of a channel, instead the Streamer token **must** be used. In addition to this, the Streamer token needs a new permission `user:edit:broadcast`.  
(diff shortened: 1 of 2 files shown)


RAnders00 (changes requested):
I would appreciate if you could also directly use the new function, i.e. make the changes to the math and lastfm modules like you wanted to originally. (https://github.com/pajbot/pajbot/issues/1118)

And then the changelog should not say that `safe_say` was added, we don't need changelog entries for things that don't change for external users.

alazymeme:
> I would appreciate if you could also directly use the new function

I can if you'd like; but the changes were more personal with the potential to use them for something else in the future....

Up to you if you still want it added.

RAnders00 (approved):
I see, I thought you were going to add those changes after this is merged with a different PR. I'll merge this one then.

alazymeme:
> I see, I thought you were going to add those changes after this is merged with a different PR. I'll merge this one then.

It might be something to add banphrase checking to all the modules anyway...since the bot really shouldn't be saying anything that's banphrased. Is that something I should do? I can make a separate pr for it when I get the chance.

RAnders00:
Would love for @pajlada to answer this. I don't really know if there's a problem with that?

alazymeme:
image

Will work on a pr

PR #1125, opened 2021-01-13
Added $(date:<timezone>) variable

zneix:
Pull request checklist:

Added a small variable `$(date:<timezone>)` outputting in format `2021-01-13`. I decided this format is the safest, as it is an ISO standard and it shouldn't be as confusing for anyone, who's using american date format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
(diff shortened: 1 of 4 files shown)


RAnders00 (approved):
Seems good! Thank you for the pull request.

PR #1132, opened 2021-01-17
Added datetime variable

alazymeme:
Pull request checklist:

My concern with adding this is: won't this make the other two variables obsolete? Since you can achieve the same thing using the datetime variable?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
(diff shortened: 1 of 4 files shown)


RAnders00 (approved):
Looks perfect! I like the extensive changelog and documentation additions. Will wait for @pajlada to review too and maybe try it out as well.

pajlada (approved):
LGTM :+1: Thanks for the contribution!

PR #1144, opened 2021-01-30
Add online and offline option to repspam module

alazymeme:
Pull request checklist:


pajlada:
```
key: enabled_by_stream_status
label: Enable moderating XXX (repetitive spam) when stream is:
default: "Online and offline"
options: ["Online only", "Offline only", "Online and offline"]

if self.settings["enabled_by_stream_status"] == "Online only" and not self.bot_is_online:
    return

if self.settings["enabled_by_stream_status"] == "Offline only" and self.bot_is_online:
    return
```

pajlada (approved):

PR #1151, opened 2021-01-31
Condense if statement in subscriber exemption in link checker module

alazymeme:
Pull request checklist:


RAnders00:
What‘s the point of this change? We still use „is True“ in a lot of other places too

alazymeme:
> What‘s the point of this change? We still use „is True“ in a lot of other places too

To remove some useless code? That's pretty much it

RAnders00:
I mean like, just in the little preview github gives me around your change i can see three more pointless "is True" statements, i just don't get why that one in particular is so outstanding and why you didnt change the other ones too.

alazymeme:
> I mean like, just in the little preview github gives me around your change i can see three more pointless "is True" statements, i just don't get why that one in particular is so outstanding and why you didnt change the other ones too.

I didn't notice the other ones. Will fix them up when I get the chance.

pajlada:
This feels like a change that will only risk messing some things up if done in mass, and have no purpose if done in one place.
Not really a change fit for a project that's "frozenish"

PR #1169, opened 2021-02-19
Add `$(datetimefromisoformat)` variable

pajlada:
Pull request checklist:

- Add `datetimefromisoformat` variable
- Add variable documentation
- Add changelog entry

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -391,6 +391,12 @@ class Bot:
         except:
             log.exception("Unhandled exception in get_date_value")
 
+    def get_datetimefromisoformat_value(self, key, extra={}):
+        try:
+            return datetime.datetime.fromisoformat(key).replace(tzinfo=datetime.timezone.utc)
+        except:
+            log.exception("Unhandled exception in get_datetimefromisoformat_value")
+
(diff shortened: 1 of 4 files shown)


pajbot/bot.py
```diff
@@ -391,6 +391,12 @@ def get_date_value(self, key, extra={}):
         except:
             log.exception("Unhandled exception in get_date_value")
 
+    def get_datetimefromisoformat_value(self, key, extra={}):
+        try:
+            return datetime.datetime.fromisoformat(key).replace(tzinfo=datetime.timezone.utc)
```
RAnders00, line 396: Should only replace the tzinfo with UTC if the parsed date has no tzinfo, since you can specify a timezone in ISO format strings as well (which we would discard, giving the wrong parsed datetime).
Maybe should also document that if no timezone is specified, UTC is assumed. (in variables.md)

RAnders00 (changes requested):

pajbot/bot.py
pajlada: Should be all good in the last two commits :ok_hand:

RAnders00 (approved):
👍🏻

PR #1170, opened 2021-02-19
Handle float values in math functions

pajlada:
Pull request checklist:

- Add a helper function to convert a number as a string to either an int or a float
- Convert math functions to use the parse_number_from_string function
- Add changelog entry for the change of math filters
- Update variable documentation for the changed math filters

### DIFF
--- a/pajbot/utils/parse_number_from_string.py
+++ b/pajbot/utils/parse_number_from_string.py
@@ -1,0 +1,18 @@
+from typing import Union
+
+
+def parse_number_from_string(var: str) -> Union[int, float]:
+    """Tries to parse a number (int or float) from the given string.
+    If no easy conversion could be made, raise a ValueError."""
+
+    try:
+        return int(var)
+    except ValueError:
+        pass
+
+    try:
+        return float(var)
+    except ValueError:
+        pass
+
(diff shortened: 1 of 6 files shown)


pajbot/utils/parse_number_from_string.py
```diff
@@ -0,0 +1,18 @@
...
+        return int(var)
+    except ValueError:
+        pass
+
+    try:
+        return float(var)
+    except ValueError:
+        pass
+
+    raise ValueError(f"could not convert string to int() with base 10 or float: {var}")
```
RAnders00, line 18:
```suggestion
    raise ValueError(f"could not convert string to number using int() or float(): {var}")
```

RAnders00 (changes requested):

pajbot/utils/parse_number_from_string.py
pajlada: Was initially trying to just use the phrasing that's provided by the int and float `ValueError` exceptions we are catching, which is
`ValueError: invalid literal for int() with base 10: 'asd'` and `ValueError: could not convert string to float: 'xd'` KKona
but this is nicer

RAnders00 (approved):

PR #1174, opened 2021-02-20
Add support for delete action in ascii module

alazymeme:
Pull request checklist:

### DIFF
--- a/pajbot/modules/ascii.py
+++ b/pajbot/modules/ascii.py
@@ -119,13 +119,12 @@ class AsciiProtectionModule(BaseModule):
             duration, punishment = self.bot.timeout_warn(
                 source, self.settings["timeout_length"], reason=self.settings["timeout_reason"]
             )
+            """ We only send a notification to the user if he has spent more than
+            one hour watching the stream. """
+            if self.settings["whisper_offenders"] and duration > 0 and source.time_in_chat_online >= timedelta(hours=1):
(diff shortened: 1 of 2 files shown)


pajbot/modules/ascii.py
```diff
@@ -33,6 +33,14 @@ class AsciiProtectionModule(BaseModule):
             default=70,
             constraints={"min_value": 20, "max_value": 1000},
         ),
+        ModuleSetting(
+            key="moderation_action",
+            label="Moderation action to apply",
+            type="options",
+            required=True,
+            default="Delete",
```
pajlada: The default behaviour should stay as "Timeout", not "Delete"

pajbot/modules/ascii.py
```diff
@@ -105,16 +113,18 @@ def on_pubmsg(self, source, message, **rest):
         if AsciiProtectionModule.check_message(message) is False:
             return
 
-        duration, punishment = self.bot.timeout_warn(
-            source, self.settings["timeout_length"], reason=self.settings["timeout_reason"]
-        )
+        if self.settings["moderation_action"] == "Delete":
+            self.bot.delete_message(tags["id"])
```
pajlada, line 117: We should return False here too to stop any further parsing of modules (could just use the same return False as below

pajlada (changes requested):

pajlada (approved):

PR #1187, opened 2021-03-07
Updated minimum required python version in docs

zneix:
Pull request checklist:

Turns out, that autobahn 21.3.1 is a part of python 3.7, so I updated documentation since we no longer support python 3.6.


pajlada (approved):
Looks good to me! :+1:

PR #1204, opened 2021-04-01
We do a little trolling (Temporary April Fools PR ¯\_(ツ)_/¯)

DatGuy1:


pajlada:
image

alazymeme:
Simon36 ☝ see the problem with this pr, is that pajlada very rarely updates his bots. So you really wouldn't get anyone with the joke.

PR #1237, opened 2021-05-09
Fix queup previous song command

ALazyMeme:
Pull request checklist:

Where previously dubtrack would provide the previous song name and details through the same endpoint, it now requires you to use the `song` endpoint with the song id in order to get the song information.

History: https://api.queup.net/room/605f18bb9ca3a90006ecf3c0/playlist/history
Song Search: https://api.queup.net/song/5668b076c6cefd710092f560


pajlada (changes requested):
The current implementation increases the API calls from 1 to 21 whenever `!previoussong` is called (not taking the cache into consideration) - this is unacceptable.

I propose `fetch_past_songs` doesn't try to fill in the `song_name` member of the `QueUpQueueSong` class, and make another method to 'hydrate' the class if `song_name` is missing, and call that 'hydrate' method from within the QueUp module instead.

Note: Not sure if hydrate is the right terminology here, but it can be changed before merging the PR.

pajlada:
@alazymeme Implementation looks good, now it would just need to be called from the module cuz right now `hydrate_song` is never used

Issue #1209, opened 2021-04-04
Idea for fixing multiple moderation modules overriding each others timeout


RAnders00:
I want to start this more as a discussion and to write down a technical idea of how we can band-aid-fix the problem that if a message is found to be bad by more than one moderation module it will apply the first module's action and not run the other modules after that. E.g. you can bypass the massping timeout by also including a banned phrase in your message.

Relevant issue: https://github.com/pajbot/pajbot/issues/1117
![image](https://user-images.githubusercontent.com/1629196/113521614-dd958d00-959a-11eb-9b93-67f7be542076.png) (braille is banned via banphrase, not ascii module in nymn's chat)

My idea was:
- Add a flag to `HandlerManager.add_handler` that is called `run_if_propagation_stopped` (or something like that). Current behaviour is to run handlers in the order they were added and to abort if one of them returns `False`. I would change it to not abort immediately, instead continue iterating handlers and only run those with `run_if_propagation_stopped=True`. Then all moderation modules would register their event handlers with this flag set to `True`.
- Add a **thread-local** state variable to `Bot` that makes `bot.timeout`, `bot.ban`, etc. behave differently if the bot is in the middle of processing a message (e.g. during [`on_pubmsg`](https://github.com/pajbot/pajbot/blob/5be9b351fcc0b3c7fdc85cf06160c43812f8525e/pajbot/bot.py#L788)).
  Instead of sending the timeout/ban command directly to the wire, the thread-local variable is updated when a module calls `bot.ban` or `bot.timeout` (for example choosing the higher of two timeouts, or a ban over a timeout). When the `on_pubmsg` handling code ends (i.e. all event handlers have been called) the action is "executed", aka sent to the wire.

We will probably need to inspect usages of `timeout` and `ban` to see whether this behaviour is not desireable for some modules, maybe we will need to add an override flag to timeout/ban methods (`immediate=True` or something like that).

Will probably affect `bot.timeout`, `bot.ban`, `bot.untimeout`, `bot.unban`.

pajlada:
This seems like a reasonable way to solve this.

PR #1253, opened 2021-05-25
Add system to always run all moderation modules and to combine their results

RAnders00:
I implemented my idea from #1209.

Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Bugfix: Users can no longer bypass long massping timeouts by also typing a banphrase in their message. (#1117, #1209, #1253)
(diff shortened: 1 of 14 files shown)


RAnders00:
I want there to be unit tests for the `ModerationActions.add` logic, either I'll do it if time and motivation allows until next weekend, if I haven't done them until now you have some stream content ideas 1x

And of course i tested it exactly once, with a combination of banphrase and massping, so more testing in general is also needed.

image

pajlada (approved):
Looks good to me! I haven't tested it out yet, but will do tomorrow together with some unit tests maybe.

Good use of types! and dataclasses look nice

pajlada:
This branch is currently running in #pajlada and #forsen

alazymeme:
@pajlada how's the testing going on this? KKonaSmile

pajlada:
Has worked well in my experience - I will merge this in without tests and we can have tests added at a later point :sunglasses:

Issue #1256, opened 2021-05-28
Add support for 7TV


TroyDota:
As the title suggests.
https://7tv.app
:)

Also the GQL implantation allows for querying of BTTV and FFZ emotes bundled into the response so you could remove the support for loading them by implementing the 7tv extension. 
If this is something you would consider adding i dont mind doing the PR.

RAnders00:
We should add 7TV support, but i‘d prefer loading ffz and bttv from those sources directly, otherwise all emote providers would be „down“ when just 7TV is down, and 7TV being relatively new doesnt exactly allow us to know of the service‘s stability.

killertofus:
In my opinion, The extension is "too niche" to spent time on implementing and maintaining it over an indefinite amount of time.

Also,adding support for one niche extension makes it harder to argue against supporting other niche extensions. Now, if the 7tv gained popularity on the levels of BTTV and FFZ, it would make sense to reconsider this at a later point in time.

pajlada:
Closed by #1260

PR #1260, opened 2021-06-05
7tv emotes support

MagicHack:
Pull request checklist:

I saw it was discussed in  #1256 that support for 7tv may be too early, so idk.
The people I host the bot for and myself use 7tv so I added it eShrug.

### DIFF
--- a/pajbot/apiwrappers/seventv.py
+++ b/pajbot/apiwrappers/seventv.py
@@ -3,7 +3,7 @@ from pajbot.apiwrappers.response_cache import ListSerializer
 from pajbot.models.emote import Emote
 
 
-class SEVENTVAPI(BaseAPI):
+class SevenTVAPI(BaseAPI):
     def __init__(self, redis):
         super().__init__(base_url="https://api.7tv.app/v2/", redis=redis)
 
--- a/pajbot/apiwrappers/seventv.py
+++ b/pajbot/apiwrappers/seventv.py
@@ -32,15 +32,30 @@ class SevenTVAPI(BaseAPI):
 
     def fetch_global_emotes(self):
         """Returns a list of global 7TV emotes in the standard Emote format."""
-        query_string = """{
-            search_emotes(query: "", globalState: "only", limit: 150, pageSize: 150) {
+        query_string = """{{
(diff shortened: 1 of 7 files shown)


pajbot/apiwrappers/seventv.py
```diff
@@ -0,0 +1,83 @@
...
+        """Returns a list of global 7TV emotes in the standard Emote format."""
+        query_string = """{
+            search_emotes(query: "", globalState: "only", limit: 150, pageSize: 150) {
+                id
+                name
+            }
+        }"""
+        params = {"query": query_string}
+        response = self.post("gql", json=params)
+        return self.parse_emotes(response["data"]["search_emotes"])
```
RAnders00, line 43: I don't like that we don't paginate here (download multiple pages if necessary)

RAnders00 (changes requested):
I only see one problem, apart from that this looks good! Thank you!

alazymeme:
Any plans to add support for the CLR modules, emote limit and I think you can add it to the quests module?

zneix:
7tv REST API is just around the corner, no idea if you want to wait for it or go with gql as it is right now.

pajbot/apiwrappers/seventv.py
MagicHack: I implemented the pagination by checking if we get a full page from the query.
There's  alse the `x-collection-size` header that could be checked for the total number of emotes, would require a post_response method to be added to the base api. Might be  a better approach.

RAnders00:
> Any plans to add support for the CLR modules, emote limit and I think you can add it to the quests module?

CLR modules support all emote providers automatically through the unified `Emote` model that all providers create.

Emote limit module also supports all emote providers.

Quests module: https://github.com/MagicHack/pajbot/blob/228700d7c601246f96bd76231a64b8f3f6733688/pajbot/modules/quests/typeemote.py#L78 We currently only select from global Twitch emotes. Support for third-party emotes in general would definitely be a different PR.

RAnders00 (approved):

RAnders00:
Please don't merge yet, waiting for @pajlada's review too

pajlada (approved):
Looks good to me - thanks for the clean addition @MagicHack

pajlada:
will merge in after #1253

PR #1261, opened 2021-06-06
Use GQL variables instead of string formatting for GQL queries

TroyDota:
Using GQL variables which avoids string format issues.

Pull request checklist:

### DIFF
--- a/pajbot/apiwrappers/seventv.py
+++ b/pajbot/apiwrappers/seventv.py
@@ -86,7 +86,7 @@ query fetchUserEmotes($id: String!) {
 }"""
 
         params = {
-            "query": query_string, 
+            "query": query_string,
             "variables": {
                 "id": channel_name,
             },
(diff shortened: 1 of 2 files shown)


pajbot/apiwrappers/seventv.py
```diff
@@ -68,18 +75,17 @@ def get_global_emotes(self, force_fetch=False):
...
+    user(id: $id) {
+        emotes {
+            id
+            name
+        }
+    }
+}"""
 
-        params = {"query": query_string}
+        params = {"query": query_string, "variables": {"id": channel_name}}
```
pajlada, line 88: can you do same formatting here as you did on line 48

pajlada (approved):
LGTM

RAnders00:
I don‘t really see any problem with the original implementation, since we‘re interpolating numbers, not arbitrary strings…

pajlada:
> I don‘t really see any problem with the original implementation, since we‘re interpolating numbers, not arbitrary strings…

Yeah safety-wise I guess we'd be fine, this seems to be more readable and standard when it comes to GQL queries that take parameters though so I'd prefer this

PR #1262, opened 2021-06-06
Fix cross-site scripting vulnerability relating to the logout route

alazymeme:
Pull request checklist:

Fixes https://github.com/pajbot/pajbot/security/code-scanning/8?query=ref%3Arefs%2Fheads%2Fmaster

### DIFF
--- a/pajbot/web/routes/base/login.py
+++ b/pajbot/web/routes/base/login.py
@@ -9,6 +9,7 @@ from flask import redirect
 from flask import render_template
 from flask import request
 from flask import session
+from flask import escape
 
 from pajbot.managers.db import DBManager
 from pajbot.managers.redis import RedisManager
@@ -174,4 +175,4 @@ def init(app):
         if return_to.startswith("/admin"):
             return_to = "/"
 
-        return redirect(return_to)
+        return redirect(escape(return_to))


RAnders00:
I don‘t think this is right. Have you tried this?

alazymeme:
> Have you tried this?

Yes, logging in and out works just fine. It returns to the page you were on and everything.

> I don‘t think this is right.

What doesn't look right about it? I did a similar pr for pajbot2: https://github.com/pajbot/pajbot2/pull/447

RAnders00:
Are you sure the argument to `redirect()` needs to be escaped? I would think it'd mess up special characters... The real vulnerability here is sending people to a URL like https://forsen.tv/logout?returnTo=http://badwebsite.com which I don't think this PR fixes, right?

RAnders00:
To elaborate on this, go to https://pajlada.pajbot.com/user/%3Chtml%3E and click Log In, and then Log Out again, you'll see you'll get sent back to the same page again. Without having tried your PR, I'd think it would send me back to https://pajlada.pajbot.com/user/%253Chtml%253E (double encoding) when using the logout.

alazymeme:
> The real vulnerability here is sending people to a URL like https://forsen.tv/logout?returnTo=http://badwebsite.com which I don't think this PR fixes, right?

I'm still working on this one. I still need to figure out how to check the host domain against a variable (since `redirectTo` starts after the host)

> To elaborate on this, go to https://pajlada.pajbot.com/user/%3Chtml%3E and click Log In, and then Log Out again, you'll see you'll get sent back to the same page again. Without having tried your PR, I'd think it would send me back to https://pajlada.pajbot.com/user/%253Chtml%253E (double encoding) when using the logout. Which is incorrect, and does not fix any security issue if I'm not mistaken.

I can see what you mean, `https://pajlada.pajbot.com/user/%3Cscript%3Ealert('XSS');%3C/script%3E` didn't seem to do anything either on the master branch - which evidently I didn't fully test and mostly relied upon CodeQL. I'll mark as invalid and continue working on the redirect one.

PR #1297, opened 2021-06-21
Migrated to v2 emote cdn endpoint

zneix:
Pull request checklist:

Using v2 paths is required for animated emotes, as can be seen in the following example.

Example of an animated emote:

    https://static-cdn.jtvnw.net/emoticons/v1/emotesv2_e0dd54510bc94631899bf64b097680a2/3.0
    https://static-cdn.jtvnw.net/emoticons/v2/emotesv2_e0dd54510bc94631899bf64b097680a2/default/dark/3.0
    https://static-cdn.jtvnw.net/emoticons/v2/emotesv2_e0dd54510bc94631899bf64b097680a2/static/dark/3.0

`default` can be replaced with `static` to always get a non-animated variant

I also moved already hardcoded emotes in the source in case v1 endpoint gets shut down later.


alazymeme:
FeelsDankMan

Issue #562, opened 2019-10-14
Check Usernames for Banphrases


Felanbird:
Adding the option to check usernames for banphrases, toggle-able option as preexisting banphrases could cause issues.

Fixes the issue that naughty names have to be manually banned but people who @ the name are timed out no problem 🏎 

streamelements example included 
![image](https://user-images.githubusercontent.com/41973452/66787296-1fb3a600-eeb1-11e9-8bd5-a0d2d424a393.png)

alazymeme:
Would this ignore the colon? What if users use /me? Will it count as a single message?

Felanbird:
> Would this ignore the colon? What if users use /me? Will it count as a single message?

This request is specific to the message sender username only, but usernames still exist as usernames even when the message is /me.

![image](https://user-images.githubusercontent.com/41973452/67538348-e3e8bf80-f6ac-11e9-8e00-5b6eaa3299ef.png)
![image](https://user-images.githubusercontent.com/41973452/67538440-0e3a7d00-f6ad-11e9-9b01-15de4ed64a63.png)
![image](https://user-images.githubusercontent.com/41973452/67538395-eea35480-f6ac-11e9-85ab-6094ebcf5c52.png)
![image](https://user-images.githubusercontent.com/41973452/67538414-fbc04380-f6ac-11e9-8c31-98b26275a6ab.png)

alazymeme:
I've started running into issues recently with a certain user(s) in a certain channel. Would be very useful, could also take out some hardcoded shit:

https://github.com/pajbot/pajbot/blob/master/pajbot/bot.py#L774-L777
https://github.com/pajbot/pajbot/blob/master/pajbot/bot.py#L794-L801

RAnders00:
Due to pajbot [being moved into maintenance mode](https://github.com/pajbot/pajbot/blob/master/README.md) I'm closing this feature request.

PR #1322, opened 2021-06-26
Remove channel specific code

ALazyMeme:
Pull request checklist:

I can understand if this was in pajlada's local versions of pajbot, but I don't see why this should be in master.

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -823,35 +823,6 @@ class Bot:
         if event.source.user == self.nickname:
             return False
 
-        if self.streamer == "forsen":
-            if "zonothene" in login:
-                self._ban(login)
-                return True
-
-            raw_m = event.arguments[0].lower()
-            if raw_m.startswith("!lastseen forsen"):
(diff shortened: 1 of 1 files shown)


pajlada:
Make an issue and we can figure out the ways to solve for this - for now this will stay

ALazyMeme:
This was an issue: https://github.com/pajbot/pajbot/issues/562
But there's no point re-opening the issue if it'll just go stale again.

ALazyMeme:
Otherwise, don't you have the option of using `UserDispatch.py` for this sort of thing?

pajlada:
Did you link the wrong issue?

ALazyMeme:
> Did you link the wrong issue? 

No? Most of the code that should be removed is in relation to that issue.

pajlada:
I'll investigate more later then, didn't read through it fully - sorry 🙏

pajlada:
This does not solve for #562 rather it removes some local code that shouldn't break anyone else's flow. I agree it's ugly, but I don't see a reason to remove it without providing an alternative solution to the *two* features you're removing.

PR #1325, opened 2021-06-30
Fix incorrect usage examples

zneix:
Pull request checklist:

Found this while browsing source out of boredom ;p


RAnders00 (approved):

RAnders00:
Shouldn't there be a changelog entry? Even though it's extremely minor I don't see on what basis we shouldn't add one.

RAnders00 (changes requested):
Shouldn't there be a changelog entry? Even though it's extremely minor I don't see on what basis we shouldn't add one.

zneix:
> Shouldn't there be a changelog entry? Even though it's extremely minor I don't see on what basis we shouldn't add one.

I just thought this is soooo minor and that nobody probably ever noticed this that realistically speaking there's no need for a changelog entry, but since this isn't Chatterino where there're releases made after every 150+ changelog entries I might as well include one.

pajlada (approved):

RAnders00 (approved):

PR #1326, opened 2021-06-30
Acknowledge users with level 2000+ as admins as well

RAnders00:
Pull request checklist:


RAnders00:
I've also seen there's some more faulty logic in that file. Once this is merged I'll fix that too.

zneix (approved):
Neat quick fix!
As for the other part of my issue, I was trying to fix something about user link colors on the user's page, however after I saw how it is handled inside `templates/user/username_link.html:14` I thought I'll give it up for now as I noticed it'd require modifying some logic inside `static/css/base.scss` which could get very messy.

ALazyMeme (approved):

zneix:
I didn't pay much attention, but the changelog entry has been mistakenly added under v1.53 section, while it should be added to Unversioned. Might be something to be fixed in another commit / PR.

PR #1355, opened 2021-07-10
Changed the timeout length limits for the Emote Limit module from [3s,1hr] to [1s,2w], and increased the maximum emote limit from 40 to 167

badoge:
Pull request checklist:

167 is the maximum number of emotes that can fit in 1 message; 2 letter emotes or emoticons


pajbot/modules/emote_limit.py
```diff
@@ -20,7 +20,7 @@ class EmoteLimitModule(BaseModule):
             required=True,
             placeholder="",
             default=15,
-            constraints={"min_value": 1, "max_value": 40},
+            constraints={"min_value": 1, "max_value": 167},
```
ALazyMeme, line 23: Is there even a reason to have a maximum value?

pajlada: maybe not, but this seems fine

pajlada (approved):

PR #1384, opened 2021-08-07
Fix pyramid parsing bug introduced in a previous PR + allow emojis in pyramids

pajlada:
- Fix an issue that could occurr with the pyramid parser in case the user posted, what seemed to us, like an empty message
- Instead of using unidecode to strip out the chatterino character, just strip on the character itself.

Pull request checklist:


ALazyMeme (approved):
ok

ALazyMeme:
I would be anal about the changelog entry wording, but it gets the message across, so I don't really care enough

PR #1400, opened 2021-08-22
Improved emote scaling in the CLR overlay

badoge:
Pull request checklist:

image-rendering was changed to pixelated in #833 but I think that auto shows better results:

Current: 
image

Updated:
image


RAnders00:
I originally went with `image-rendering: pixelated` because of low-res emotes looking worse (in my opinion) when they are scaled up:

image
image

image
image

image
image

badoge:
I changed it so that it uses pixelated when the emote is shorter than 100px and auto otherwise.

But I found another problem related to pixel art emotes FeelsDankMan
auto (after update):
image
pixelated (current):
image

pajlada (approved):
Looks good to me

PR #1401, opened 2021-08-22
Hide scrollbar in CLR overlay

badoge:
Pull request checklist:

It seems that #833 didn't fix the issue ¯\\\_(ツ)_/¯

Current:
C9C0TY2pbC
Updated:
VEIB0QFaJ5


RAnders00:
The fix in #833 was only for `!#showemote` emotes going out-of-screen showing the scrollbar, as far as I remember. Thanks for looking into this bug too.

ALazyMeme (approved):

PR #1403, opened 2021-08-24
Fix incorrect grammar on bot homepage

ALazyMeme:
Pull request checklist:

Proper English has it that:

A word ending in s that requires an apostrophe - most commonly due to ownership, should end with the apostrophe as opposed to `'s`. This addresses that issue. Refer below:

After changes:
image

Before changes:
image

Note that `streamer.name` is the correct capitalization, whereas `streamer.full_name` is in lowercase. I've used `streamer.full_name` in order to check for the trailing s and printed with `streamer.name` as the final result.

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -6,6 +6,7 @@
 {% set streamer_name = streamer.full_name' %}
 {% else %}
 {% set streamer_name = streamer.full_name's%}
+{% endif %}
 <h4>
     This is the website for <strong>{{ bot.name }}</strong> that moderates <a href="http://twitch.tv/{{ streamer.full_name }}">{{ streamer_name }}</a> channel.
 </h4>
--- a/templates/home.html
+++ b/templates/home.html
@@ -3,9 +3,9 @@
 {% block title %}Home{% endblock %}
 {% block body %}
 {% if streamer.full_name.endswith('s') %}
-{% set streamer_name = streamer.full_name' %}
+{% set streamer_name = streamer.full_name + "'" %}
 {% else %}
-{% set streamer_name = streamer.full_name's%}
+{% set streamer_name = streamer.full_name + "'s" %}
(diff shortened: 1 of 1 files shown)


RAnders00 (changes requested):
You can do this in a simpler way by just saying `{ streamer.name }'{ "" if name_ends_in_s else "s" }`

pajlada (approved):

templates/home.html
```diff
@@ -3,7 +3,7 @@
 {% block title %}Home{% endblock %}
 {% block body %}
 <h4>
-    This is the website for <strong>{{ bot.name }}</strong> that moderates <a href="http://twitch.tv/{{ streamer.full_name }}">{{ streamer.name }}'s</a> channel.
+    This is the website for <strong>{{ bot.name }}</strong> that moderates <a href="http://twitch.tv/{{ streamer.full_name }}">{{ streamer.name }}'{{% "" if streamer.name.endswith('s') else "s" %}}</a> channel.
```
RAnders00, line 6:
```suggestion
    This is the website for <strong>{{ bot.name }}</strong> that moderates <a href="http://twitch.tv/{{ streamer.full_name }}">{{ streamer.name }}'{{% "" if lower(streamer.name).endswith('s') else "s" %}}</a> channel.
```

RAnders00 (changes requested):

templates/home.html
ALazyMeme: Wouldn't it be best to use `streamer.full_name`, since that pulls it from the main part of the config as opposed to `streamer.name` that uses the streamer's display name?

ALazyMeme: Actually, we have to use `streamer.full_name`, else we could be using asian display names and error out as a result.

RAnders00 (approved):

PR #1409, opened 2021-08-26
Update rate limits according to new Twitch documentation

ALazyMeme:
Pull request checklist:

Adjusted whisper rate limits to assign with Twitch's new limits. While old verified bots still have the old whisper limits, unless you're sending that many whispers (for a single channel bot), you shouldn't notice a difference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 
 - Bugfix: CLR overlay scrollbar is now hidden. (#1401)
 - Fix: Don't allow following of empty Twitter usernames through the admin page. (#1395)
+- Dev: Updated whisper rate limits to align with Twitch specification of: 3 per second, up to 100 per minute. (#1409)
 
 ## v1.54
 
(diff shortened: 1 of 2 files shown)


pajbot/managers/irc.py
```diff
@@ -129,8 +129,8 @@ def _can_send(self, is_whisper):
         if is_whisper:
             return (
                 self.num_privmsg_sent < self.bot.tmi_rate_limits.privmsg_per_30
-                and self.num_whispers_sent_second < self.bot.tmi_rate_limits.whispers_per_second
-                and self.num_whispers_sent_minute < self.bot.tmi_rate_limits.whispers_per_minute
+                and self.num_whispers_sent_second < 2
+                and self.num_whispers_sent_minute < 90
```
pajlada, line 133: These values should still be stored as a variable. They make sense to be in the `tmi_rate_limits` object.

pajlada (changes requested):

pajbot/managers/irc.py
ALazyMeme: What is the point of storing them as variables if the rate limit is set for all accounts? Doesn't it seem a bit redundant? Since the `num_whispers_sent_second` and `num_whispers_sent_minute` variables are not referenced anywhere else?

pajlada: Don't introduce new "magic numbers" into the code that will be harder to find and modify in the future

https://www.perforce.com/resources/qac/high-integrity-cpp-coding-standard-expressions

pajlada (approved):

PR #1413, opened 2021-08-28
Update max chunk processing value in pnsl module

ALazyMeme:
Pull request checklist:

According to this: https://dev.twitch.tv/docs/irc/guide#rate-limits
The new limits are 100/30s per channel for all users.

### DIFF
--- a/pajbot/modules/pnsl.py
+++ b/pajbot/modules/pnsl.py
@@ -33,7 +33,7 @@ class PNSLModule(BaseModule):
             required=True,
             placeholder="",
             default=30,
-            constraints={"min_value": 1, "max_value": 500},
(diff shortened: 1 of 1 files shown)


ALazyMeme:
@pajlada Should I also update the `chunk_delay` to have a `min_value` of 30 so as not to allow users to cross the limit?

Felanbird:
This PR just changes stuff just to change stuff.

1. The documented limitation is not true, test it yourself.
2. You are forcing limitations on grandfathered verified bots, see 1.
3. This is not a publicly used module, if you know what it is you know how to properly use it, we probably shouldn't even be forcing a 500 max but the bot is ran 1 connection anyway and would never be able to get 500 bans in in 30 seconds anyway.

PR #1417, opened 2021-08-29
Added the ability to change the bot response method in the clip module

ALazyMeme:
Pull request checklist:


ALazyMeme:
Should I be making a changelog entry for this? The default values don't change the behaviour of the module compared to what it was before.

pajlada (approved):

PR #1423, opened 2021-08-29
Added the ability to change the bot response method in the wolfram module

ALazyMeme:
Pull request checklist:

- [x] Tested and confirmed working

### DIFF
--- a/pajbot/modules/wolfram.py
+++ b/pajbot/modules/wolfram.py
@@ -199,7 +199,7 @@ class WolframModule(BaseModule):
             level=self.settings["level"],
             description="Ask Wolfram Alpha a question",
             command="query",
-            can_execute_with_whisper=bool(self.settings["response_method"] == "reply"),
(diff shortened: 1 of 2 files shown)


pajbot/modules/wolfram.py
```diff
@@ -193,6 +199,7 @@ def load_commands(self, **options):
             level=self.settings["level"],
             description="Ask Wolfram Alpha a question",
             command="query",
+            can_execute_with_whisper=bool(self.settings["response_method"] == "reply"),
```
RAnders00, line 202: I don't think you need the `bool()` here, last time I checked the result of `==` was always a bool.

RAnders00 (dismissed):

pajbot/modules/wolfram.py
ALazyMeme: So I just enclose it in brackets?

pajlada (approved):

PR #1439, opened 2021-09-09
Don't lowercase first letter in `send_message_to_user` function

ALazyMeme:
Pull request checklist:

The username is the salutation, therefore the first letter should be capitalized as per proper English. **Should** be, but we shouldn't enforce it, but rather display the message as entered by the user/module.

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -659,14 +659,14 @@ class Bot:
 
     def send_message_to_user(self, user, message, event, method="say"):
         if method == "say":
-            self.say(f"@{user.name}, {lowercase_first_letter(message)}")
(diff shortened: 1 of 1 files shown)


ALazyMeme:
Also I think that regardless of if it's proper English or not, the decision should really be made for the person making the commands/formatting the module. We shouldn't enforce this sort of a thing on people.

RAnders00 (dismissed):
https://github.com/pajbot/pajbot/blob/cd307f392600770bb97db0bdd82ff663ede847e5/pajbot/bot.py#L1053-L1054 You should remove the function as well since it would be unused after this change.

PR #1442, opened 2021-09-10
Add support for twitch reply threads

ALazyMeme:
Pull request checklist:

Thanks @DatGuy1 for the code

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -666,11 +666,28 @@ class Bot:
             self.me(message)
         elif method == "reply":
             if event.type in ["action", "pubmsg"]:
-                self.say(f"@{user.name}, {lowercase_first_letter(message)}")
+                msg_id = next(tag["value"] for tag in event.tags if tag["key"] == "id")
+                self.reply(msg_id, message)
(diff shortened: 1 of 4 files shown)


pajbot/bot.py
```diff
@@ -666,12 +666,29 @@ def send_message_to_user(self, user, message, event, method="say"):
...
             if event.type in ["action", "pubmsg"]:
-                self.say(f"@{user.name}, {message}")
+                msg_id = next(tag["value"] for tag in event.tags if tag["key"] == "id")
+                self.reply(msg_id, message)
             elif event.type == "whisper":
                 self.whisper(user, message)
         else:
             log.warning("Unknown send_message method: %s", method)
 
+    def reply(self, msg_id, message, channel=None):
```
pajlada, line 676: I have tested that replying still works even if msg_id is empty or msg_id is invalid (e.g. `xD`), it just won't reply.
Since it doesn't reply, and doesn't add the username at the front, it's important that we run the `utils.clean_up_message` function (which you do) :+1:

pajlada (approved):

PR #1443, opened 2021-09-11
Updated the broadcast manage scope to match updated helix docs

ALazyMeme:
Pull request checklist:

It was updated: https://dev.twitch.tv/docs/api/reference#modify-channel-information
The old scope remains valid, but is only officially used for extensions, so better off updating it


ALazyMeme:
Yes this works

pajlada (approved):

PR #1447, opened 2021-09-11
Revert bot response changes in #1418

ALazyMeme:
Pull request checklist:

I think I got ahead of myself here. Not only does this not make sense for the emotes module, but also I believe it may have broken the module, as only one message of emotes is posted:

image


pajlada (approved):

ALazyMeme:
Yes, just tested and this definitely did break the module.

PR #1456, opened 2021-09-12
Add option to send whispers to control hub

MagicHack:
Pull request checklist:

Added the option  `whisper_output_mode = control` to output whispers to control hub since whispers don't work for "new" bots and outputting in the streamer's chat can be annoying.

(Maybe the control_hub channel could be an attribute of the bot class instead of getting it in the config every time we need it...)

### DIFF
--- a/configs/example.ini
+++ b/configs/example.ini
@@ -28,7 +28,7 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; `whisper_output_mode = normal` sends whispers as normal. (This is the default if you don't configure this option)
 ; `whisper_output_mode = disabled` will make it so the bot does not send any whispers. Outgoing whispers will be dropped.
 ; `whisper_output_mode = chat` outputs all messages intended to be whispered to the streamer's chat instead of whispering them.
-; `whisper_output_mode = control` outputs all messages intended to be whispered to the control hub chat instead of whispering them.
+; `whisper_output_mode = control_hub` outputs all messages intended to be whispered to the control hub chat instead of whispering them.
 ;whisper_output_mode = normal
 
 ; Optional section if you want to make the "Wolfram Alpha Query" module available for use:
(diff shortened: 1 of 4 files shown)


configs/example.ini
```diff
@@ -28,6 +28,7 @@ db = postgresql+psycopg2:///pajbot?options=-c%%20search_path%%3Dpajbot1_streamer
 ; `whisper_output_mode = normal` sends whispers as normal. (This is the default if you don't configure this option)
 ; `whisper_output_mode = disabled` will make it so the bot does not send any whispers. Outgoing whispers will be dropped.
 ; `whisper_output_mode = chat` outputs all messages intended to be whispered to the streamer's chat instead of whispering them.
+; `whisper_output_mode = control` outputs all messages intended to be whispered to the control hub chat instead of whispering them.
```
RAnders00, line 31: I think I would prefer if it was called `control_hub` instead of `control`

pajbot/bot.py
```diff
@@ -646,6 +646,11 @@ def whisper(self, user, message):
             self.irc.whisper(user.login, message)
         if self.whisper_output_mode == WhisperOutputMode.CHAT:
             self.privmsg(f"{user}, {message}")
+        if self.whisper_output_mode == WhisperOutputMode.CONTROL:
+            chub = self.config["main"].get("control_hub", None)
+            if chub is not None:
+                chub = f"#{chub}"
+            self.privmsg(f"{user}, {message}", chub)
```
RAnders00, line 653: What happens here if there is no `control_hub` configured in the config?

RAnders00 (changes requested):

pajbot/bot.py
MagicHack: privmsg is called with None as target channel, so it sends the message to the streamers channel by default. Could throw an error instead eShrug

ALazyMeme:
I don't see a good enough reason to merge this. It's the equivalent of using the `disabled` option, but it's just chewing up rate limits. Could you please provide a use case where this might be useable?

MagicHack:
> I don't see a good enough reason to merge this. It's the equivalent of using the `disabled` option, but it's just chewing up rate limits. Could you please provide a use case where this might be usable?

It's mostly for debugging purpose since a lot of moderator/admin commands whisper the information by default. For the viewers it's effectively as if it was disabled, but for the admin/mods it provides the reply to commands without spamming the main channel. Of course if all the banphrases are notifying the users it's gonna waste of lot of the rate limit.

pajbot/bot.py
RAnders00: You should probably do nothing if no control hub is configured and print a warning on the console along the lines of ``Whisper output mode set to `control_hub` but no control hub configured in config, following whisper will not be sent: To {user}: {message}``

RAnders00 (changes requested):

pajlada (approved):

PR #1469, opened 2021-09-19
Switch to `regex` for banphrase regexes

### LINKED ISSUE #1468: Consider migrating from "re" to "regex" library for matching regular expressions
After my small research I've found out that in Python there is a standard `re` library for matching regular expressions, but there's also a [`regex` library](https://pypi.org/project/regex), which provides some functionality that the `re` library is missing.
I'd personally love to use some features that are not in `re`, but `regex` supports, one of them being [`(?&name)`](https://bitbucket.org/mrabarnett/mrab-regex/issues/152).

This is [one of regexes (not safe for stream)](https://regex101.com/r/1w4rfa/5) with which I've ran into an issue - it works fine with PCRE/PCRE2 flavours and it should work with `regex` library (given its readme).

RAnders00:
Pull request checklist:

Fixes #1468

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
-- Minor: Regex banphrases now support more complex regex features. (#1468, #1469)
+- Minor: Regex banphrases now support more complex regex features. (#1469)
 - Bugfix: Made `!debug playsound` use default cooldown values (5s global / 15s user). (#1474)
 
 ## v1.55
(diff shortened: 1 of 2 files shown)


zneix:
Did a little bit of testing on my own:
1. Some basic banphrases without regex weren't affected and seem to work fine.
2. Baisc regex banphrases with what `re` has been supporting so far works as well. I tested this with [banphrases](https://haste.zneix.eu/raw/zonulohafe) on [my own pajbot instance](https://twitchbot.zneix.eu)
3. Banphrase regex mentioned in #1468 does work with this change.
4. I'm unsure whether some other advanced regexes, in e.g. #forsen would break. Maybe @leppunen or pajlada can help with this one.

pajlada (approved):
Tested on a bunch of production banphrases, and things seem to work as expected. LGTM :+1:

PR #1502, opened 2021-10-05
Added `--disable` and `--enable` command arguments

zneix:
Pull request checklist:

These new arguments can be used within `!edit command` and/or `!add command` to enable/disable new/existing commands.
Also corrected the exception message since it wasn't technically correct before.


ALazyMeme:
Does disabling them actually work? Or do they still just disappear from the list?

Felanbird:
> 
> 
> Does disabling them actually work? Or do they still just disappear from the list?

This does not fix the issue of disabled commands not showing up, but it does give more reason to actually fix the underlying issue.

PR #1580, opened 2021-10-31
Add streamer config fallback

ALazyMeme:
Pull request checklist:

Should the bot admin not input a `streamer` in the config file, the bot will attempt to use the `streamer_display` config option. This also means that it is no longer required to input the `streamer` in the config file (providing there's no Asian characters in the display name.

It'd be worth just moving to IDs at some point in order to simplify the process.


pajlada:
I think just moving to IDs (or allowing a user to use IDs) will be better than this half-solution.

PR #1590, opened 2021-11-06
Allow configuring streamer, bot, admin, and control hub using Twitch User IDs instead of Twitch User Logins

ALazyMeme:
Pull request checklist:


pajbot/config.py
```diff
@@ -1,13 +1,17 @@
...
+from pajbot.apiwrappers.twitch.helix import TwitchHelixAPI
 
 ConfigSection = Dict[str, str]
 Config = Dict[str, ConfigSection]
 
 
 def load_streamer_and_channel(config: Config) -> Tuple[str, str]:
-    if "streamer" in config["main"]:
+    if "streamer_id" in config["main"]:
+        streamer = TwitchHelixAPI.get_login(config["main"]["streamer_id"])
```
pajlada, line 12: No API requests should take place in this file

configs/example.ini
```diff
@@ -1,13 +1,13 @@
...
-; Twitch login name of the channel the bot runs in
-streamer = streamer_name
-; login name of the primary admin (will be granted level 2000 initially)
-admin = your_own_name
-; an additional channel the bot will join and receive commands from. (login name)
+; Twitch ID of the bot account
+bot_id = 123456789
+; Twitch ID of the channel the bot runs in
+streamer_id = 123456789
+; Login ID of the primary admin (will be granted level 2000 initially)
```
pajlada, line 6:
```suggestion
; Twitch User ID of the primary admin (will be granted level 2000 initially)
```

configs/example.ini
```diff
@@ -1,13 +1,13 @@
...
-; Twitch login name of the bot account
-nickname = the_bot_name
-; Twitch login name of the channel the bot runs in
-streamer = streamer_name
-; login name of the primary admin (will be granted level 2000 initially)
-admin = your_own_name
-; an additional channel the bot will join and receive commands from. (login name)
+; Twitch ID of the bot account
+bot_id = 123456789
+; Twitch ID of the channel the bot runs in
```
pajlada, line 4:
```suggestion
; Twitch User ID of the channel the bot runs in
```

configs/example.ini
```diff
@@ -1,13 +1,13 @@
 [main]
-; Twitch login name of the bot account
-nickname = the_bot_name
-; Twitch login name of the channel the bot runs in
-streamer = streamer_name
-; login name of the primary admin (will be granted level 2000 initially)
-admin = your_own_name
-; an additional channel the bot will join and receive commands from. (login name)
+; Twitch ID of the bot account
```
pajlada, line 2:
```suggestion
; Twitch User ID of the bot account
```

configs/example.ini
```diff
@@ -1,13 +1,13 @@
...
+bot_id = 123456789
+; Twitch ID of the channel the bot runs in
+streamer_id = 123456789
+; Login ID of the primary admin (will be granted level 2000 initially)
+admin_id = 123456789
+; an additional channel the bot will join and receive commands from. (user id)
 ; You can remove this config entry if you don't want to use this feature
```
pajlada, line 9:
```suggestion
; Twitch User ID of an additional channel the bot will join and receive commands from.
; You can remove this config entry if you don't want to use this feature
```

pajbot/bot.py
```diff
@@ -130,16 +120,44 @@ def __init__(self, config: cfg.Config, args: argparse.Namespace) -> None:
...
         self.twitch_helix_api: TwitchHelixAPI = TwitchHelixAPI(RedisManager.get(), self.app_token_manager)
 
-        self.bot_user_id = self.twitch_helix_api.get_user_id(self.nickname)
+        # bot
+        if "bot_id" in config["main"]:
+            self.bot_user_id = config["main"]["bot_id"]
+            self.nickname = self.twitch_helix_api.get_login(self.bot_user_id)
+        elif "nickname" in config["main"]:
+            self.nickname = config["main"].get("nickname", "pajbot")
+            self.bot_user_id = self.twitch_helix_api.get_user_id(self.nickname)
```
pajlada, line 129: if bot_id and nickname is missing, we silently fail and will error out in a weird way later

pajbot/bot.py
```diff
@@ -130,16 +120,44 @@ def __init__(self, config: cfg.Config, args: argparse.Namespace) -> None:
...
+        if "streamer_name" in config["web"]:
+            self.streamer_display = config["web"]["streamer_name"]
+        elif "streamer_name" not in config["web"]:
+            self.streamer_display = self.twitch_helix_api.get_display_name(self.streamer_user_id)
+
+        # control hub
+        if "control_hub_id" in config["main"]:
+            self.control_hub: Optional[str] = self.twitch_helix_api.get_login(config["main"]["control_hub_id"])
+        else:
+            self.control_hub: Optional[str] = config["main"].get("control_hub", None)
```
pajlada, line 151: The control_hub variable can be defined before we start checking if it exists in config["main"]
`self.control_hub: Optional[str]`

pajbot/web/__init__.py
```diff
@@ -147,13 +164,13 @@ def init(args):
     default_variables = {
         "version": version,
         "last_commit": last_commit,
-        "bot": {"name": config["main"]["nickname"]},
+        "bot": bot_login,
```
pajlada, line 167: This is not the same, `default_variables["bot"]` used to be a dictionary with the key `name` with the bot's login name. Now `default_variables["bot"]` is  the bot's login name.

pajlada (dismissed):

pajlada:
To document it in writing as well: I'm ok with this upgrade, but I need it to be backwards-compatible. All old ways of configuring the bot must continue working.

pajbot/config.py
RAnders00: Also, this is not going to work, TwitchHelixAPI.get_login needs a `self` parameter so it needs to be called on an instance of the API

pajbot/bot.py
RAnders00: You should use `require_user_id` which already exists in the helix api wrapper: https://github.com/pajbot/pajbot/blob/e28d12cc430ff7567208a439d1c88a5204ef6378/pajbot/apiwrappers/twitch/helix.py#L223-L227

pajbot/bot.py
```diff
@@ -130,16 +120,44 @@ def __init__(self, config: cfg.Config, args: argparse.Namespace) -> None:
...
+            self.bot_user_id = config["main"]["bot_id"]
+            self.nickname = self.twitch_helix_api.get_login(self.bot_user_id)
+        elif "nickname" in config["main"]:
+            self.nickname = config["main"].get("nickname", "pajbot")
+            self.bot_user_id = self.twitch_helix_api.get_user_id(self.nickname)
+
+        # streamer
+        if "streamer_id" in config["main"]:
+            self.streamer_user_id = config["main"]["streamer_id"]
+            self.streamer = self.broadcaster = self.twitch_helix_api.get_login(self.streamer_user_id)
```
RAnders00, line 134: You should create a function `require_login`, similar to `require_user_id` and then use it here and in all other places where you don't want `None` to be returned

ALazyMeme: Is it worth making the same thing for the display name?

pajlada:
this is now in a testable state, I will deploy it to my prod system and test it out

ALazyMeme:
`$(tb:broadcaster)` doesn't seem to work anymore:
image

image

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Minor: Allow configuring streamer, bot, admin, and control hub using Twitch User IDs instead of Twitch User Logins. (#1590)
```
RAnders00, line 7:
```suggestion
- Minor: Allow configuring streamer, bot, admin, and control hub using Twitch User IDs instead of Twitch User Logins. **Specifying user logins in the config has therefore now been deprecated and may be stop working in a future release.** (#1590)
```

configs/example.ini
```diff
@@ -1,13 +1,13 @@
...
+; Twitch User ID of the bot account
+bot_id = 123456789
+; Twitch User ID of the channel the bot runs in
+streamer_id = 123456789
+; Twitch User ID of the primary admin (will be granted level 2000 initially)
+admin_id = 123456789
+; Twitch User ID of an additional channel the bot will join and receive commands from.
 ; You can remove this config entry if you don't want to use this feature
-control_hub = a_channel_for_all_your_bots_to_join
+control_hub_id = 123456789
```
RAnders00, line 10: Further down in the config there is a `streamer_name` option under `[web]`, shouldn't that be removed/replaced as well?

pajbot/bot.py
```diff
@@ -130,21 +136,31 @@ def __init__(self, config: cfg.Config, args: argparse.Namespace) -> None:
...
-        self.broadcaster = self.twitch_helix_api.get_user_basics_by_login(self.streamer)
-        if self.broadcaster is None:
-            raise ValueError("The streamer login name you entered under [main] does not exist on twitch.")
+        self.bot_user: UserBasics = cfg.load_bot(config, self.twitch_helix_api)
 
-        self.streamer_user_id = self.broadcaster.id
+        self.control_hub_user: Optional[UserBasics] = self._load_control_hub(config)
+        self.control_hub_channel: Optional[str]
+        if self.control_hub_user:
+            self.control_hub_channel = f"#{self.control_hub_user.login}"
```
RAnders00, line 152: shouldn't there be a `else: self.control_hub_channel = None` here since otherwise it'll be unitialized and usages of `bot.control_hub_channel` will result in a `KeyError`?

Regardless of that, might it not be cleaner to go with just `self.control_hub` and ditch `self.control_hub_channel` (to make it consistent with `self.streamer` etc.)

RAnders00 (dismissed):

ALazyMeme:
Based on my logs, there might be some other ones not working:
`pajbot@weest[841]: [2021-11-21 09:22:38,809] [INFO   ] Set current stream chunk here to <pajbot.models.stream.StreamChunk object at 0x7f4525753d00>`

CHANGELOG.md
ALazyMeme: I'm not sure if I agree with this. I mean we haven't removed the password in config file yet have we? We still just spit a deprecation warning.

pajbot/config.py
```diff
@@ -1,20 +1,117 @@
...
+
+    if "control_hub_id" in config["main"]:
+        return config["main"]["control_hub_id"], None
+    elif "control_hub" in config["main"]:
+        log.warning(
+            "DEPRECATED - Specify the control hub's Twitch User ID using control_hub_id instead of their login name in control_hub"
+        )
+        return None, config["main"]["control_hub"]
+
+    raise KeyError("Missing control_hub_id key from config")
```
ALazyMeme, line 95: This wasn't a required field before. Certain channels I don't have a control hub set and I don't think it should be required.

CHANGELOG.md
RAnders00: Well, that's what deprecated means - still works but we plan on removing it in the future.

pajbot/config.py
pajlada: Fixed in [`5e8f8e3` (#1590)](https://github.com/pajbot/pajbot/pull/1590/commits/5e8f8e335eeafa6bd1aa125f956bbf443c4c44b7)

pajlada:
> Based on my logs, there might be some other ones not working: `pajbot@weest[841]: [2021-11-21 09:22:38,809] [INFO ] Set current stream chunk here to <pajbot.models.stream.StreamChunk object at 0x7f4525753d00>`

Exactly the same in master: `[2021-11-21 12:34:55,719] [INFO   ] Set current stream chunk here to <pajbot.models.stream.StreamChunk object at 0x7f2d225e5c40>`

this is the original debug output

pajbot/bot.py
pajlada: You're correcl, I thought `a: Optional[str]` would be default initialized in some way but that's not the case. I've replaced it with `a: Optional[str] = None` instead of the else case, so we don't have to specify the typing in multiple places

I'll keep `self.control_hub_channel` (it's consistent with `self.channel` to not have to do the logic in the two places where whispers take place and in the quit logic

pajlada (approved):

PR #1638, opened 2021-12-13
Added an error for title length.

mmattbtw:
Pull request checklist:

I did this because if a title is too long, pajbot responds saying "Banned words in title" or something along the lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 
 - Minor: The `redirect_uri` config option under the `[twitchapi]` section is now optional with a sane default value: `https://<domain>/login/authorized`. (#1618)
 - Minor: Allow configuring streamer, bot, admin, and control hub using Twitch User IDs instead of Twitch User Logins. **Specifying user logins in the config has therefore now been deprecated and may stop working in a future release.** (#1590)
-- Bugfix: Added an error for a title being too long (#1638)
+- Bugfix: Added an error for a title being too long. (#1638)
 - Bugfix: Fix tweet-manager streaming if bot follows non-existent twitter users. (#1589)
 - Dev: Replaced Flask-Scrypt with the built-in secrets module to generate the Flask secret key. (#1613)
 - Dev: The flask secret key is now stored in redis instead of in the config.ini file. (#1613)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 
 - Minor: The `redirect_uri` config option under the `[twitchapi]` section is now optional with a sane default value: `https://<domain>/login/authorized`. (#1618)
 - Minor: Allow configuring streamer, bot, admin, and control hub using Twitch User IDs instead of Twitch User Logins. **Specifying user logins in the config has therefore now been deprecated and may stop working in a future release.** (#1590)
+- Minor: Added an error for a title being too long (#1638)
```
ALazyMeme, line 9: This should be a bugfix entry

RAnders00 (approved):

RAnders00:
Is this 140 length documented somewhere? Or did you determine it empirically?

mmattbtw:
@RAnders00 If you go into the Twitch dashboard and set your title, there is a character counter

ALazyMeme (approved):

RAnders00:
What happens when you try and set a too-long title currently? I‘m mostly asking because if the API returns a specific error code (in the returned HTTP response body) indicating that the title was too long then we wouldn‘t need to hardcode the length of 140.

mmattbtw:
@RAnders00 this is what currently happens.
image
Handled here: https://github.com/pajbot/pajbot/blob/6e59fd6e5eda317f7e9d862d226de8c06141774f/pajbot/modules/basic/stream_update.py#L138-L140

I looked into the Twitch documentation, and for the response codes it doesn't really get detailed, so I wouldn't know

ALazyMeme:
> What happens when you try and set a too-long title currently? I‘m mostly asking because if the API returns a specific error code (in the returned HTTP response body) indicating that the title was too long then we wouldn‘t need to hardcode the length of 140.

Would the hardcoding of the limit not be better? Therefore we are not wasting API calls trying to set a title?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 
 - Minor: The `redirect_uri` config option under the `[twitchapi]` section is now optional with a sane default value: `https://<domain>/login/authorized`. (#1618)
 - Minor: Allow configuring streamer, bot, admin, and control hub using Twitch User IDs instead of Twitch User Logins. **Specifying user logins in the config has therefore now been deprecated and may stop working in a future release.** (#1590)
+- Bugfix: Added an error for a title being too long (#1638)
```
pajlada, line 9:
```suggestion
- Bugfix: Added an error for a title being too long. (#1638)
```

pajlada (approved):
Since we can't differentiate the 400 status code between the various errors, doing a client-side check like this makes sense to me :+1:

PR #1669, opened 2021-12-30
Fix the bot downgrading bot scopes when logging in on the normal `/login` page using the bot account

ALazyMeme:
Pull request checklist:

We have protections in place for the streamer's scopes being overriden. I've come across an issue recently where a user had logged into the web panel using the bot account - thus resetting the scopes. This aims to fix anything accidental like this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 
 - Minor: The `redirect_uri` config option under the `[twitchapi]` section is now optional with a sane default value: `https://<domain>/login/authorized`. (#1618, #1666)
 - Minor: Allow configuring streamer, bot, admin, and control hub using Twitch User IDs instead of Twitch User Logins. **Specifying user logins in the config has therefore now been deprecated and may stop working in a future release.** (#1590)
+- Bugfix: Fix the bot downgrading bot scopes when logging in on the normal `/login` page using the bot account. (#1669)
 - Bugfix: Added an error for a title being too long. (#1638)
 - Bugfix: Fix tweet-manager streaming if bot follows non-existent twitter users. (#1589)
 - Dev: Replaced Flask-Scrypt with the built-in secrets module to generate the Flask secret key. (#1613)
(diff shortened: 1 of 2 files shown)


RAnders00 (dismissed):
I‘d like to have the code duplication removed, can you try and combine both streamer and bot logic into a single block?

ALazyMeme:
> I‘d like to have the code duplication removed, can you try and combine both streamer and bot logic into a single block?

Aside from putting it all into a single if statement (and there would still be code duplication), I can't see a way of combining it. I looked at `if me.login in (app.bot_user.login.lower(), app.streamer.login.lower())`, but I'm not sure how you would separate the functions that are different (namely the logging).

```python
        # bot login
        if me.login == app.bot_user.login.lower():
            # There's a possibility the owner of the bot account will login using the normal login button.
            # We only want to update their access token if the returned scope contains the special scopes requested
            # in /bot_login
            if set(access_token.scope) < set(bot_scopes):
                log.info("Bot account logged in but not all scopes present, will not update bot token")
            else:
                redis = RedisManager.get()
                redis.set(f"authentication:user-access-token:{me.id}", json.dumps(access_token.jsonify()))
                log.info("Successfully updated bot token in redis")
        # streamer login
        elif me.login == app.streamer.login.lower():
            # there's a good chance the streamer will later log in using the normal login button.
            # we only update their access token if the returned scope containes the special scopes requested
            # in /streamer_login

            # We use < to say "if the granted scope is a proper subset of the required scopes", this can be case
            # for example when the bot is running in its own channel and you use /bot_login,
            # then the granted scopes will be a superset of the scopes needed for the streamer.
            # By doing this, both the streamer and bot token will be set if you complete /bot_login with the bot
            # account, and if the bot is running in its own channel.
            if set(access_token.scope) < set(streamer_scopes):
                log.info("Streamer logged in but not all scopes present, will not update streamer token")
            else:
                redis = RedisManager.get()
                redis.set(f"authentication:user-access-token:{me.id}", json.dumps(access_token.jsonify()))
                log.info("Successfully updated streamer token in redis")
```

pajlada:
Tested this functionality on my dev setup by logging into the streamer/bot login and normal login with both accounts, works as I would expect. Just missing a changelog entry now and it's good to merge.

I consolidated the bot/streamer token checks. Running a self-bot is even less supported but still possible if you try hard enough.

pajlada (approved):

PR #1722, opened 2022-01-26
Add a self timeout module

ALazyMeme:
Pull request checklist:

Added a selftimeout module, allowing users to timeout themselves for a random amount of time

Thanks @tetyys for the inspiration

### DIFF
--- a/pajbot/modules/basic/roll.py
+++ b/pajbot/modules/basic/roll.py
@@ -150,7 +150,7 @@ class RollModule(BaseModule):
                 if timeout_length > 1209600:
                     timeout_length = 1209600
 
-                self.bot.timeout(source, timeout_length, f"You rolled a {rolled_value}!")
+                self.bot.timeout(source, timeout_length, f"You rolled a {rolled_value}!", once=True)
 
                 if self.settings["announce_rolls"] is True:
                     self.bot.send_message_to_user(source, f"You rolled a {rolled_value}!", event, method="reply")
--- a/pajbot/modules/basic/roll.py
+++ b/pajbot/modules/basic/roll.py
@@ -22,6 +22,13 @@ class RollModule(BaseModule):
             default=False,
         ),
         ModuleSetting(
+            key="vip_only",
+            label="Only allow VIPs to use the !roll command.",
+            type="boolean",
+            required=True,
+            default=False,
+        ),
+        ModuleSetting(
             key="global_cd",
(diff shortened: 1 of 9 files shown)


TETYYS:
> using .timeout forsen 20160m literally doesn't work for some reason

I experienced that too when using just `123s`, maybe they removed support for s/m/h/w?

ALazyMeme:
> maybe they removed support for s/m/h/w?

Don't think so, since `.timeout forsen 1w` works just fine. Maybe converting such a high value is too much work for Twitch's servers Kappa

pajbot/modules/basic/roll.py
```diff
@@ -0,0 +1,158 @@
...
+                self.bot.send_message_to_user(
+                    source, f"You rolled a {rolled_value}! {self.settings['zero_response']}", event, method="reply"
+                )
+            else:
+                timeout_length = self.seconds_conversion(rolled_value)
+                # Check if timeout value is over Twitch's maximum
+                if timeout_length > 1209600:
+                    timeout_length = 1209600
+
+                self.bot.timeout(source, timeout_length, f"You rolled a {rolled_value}!")
```
TETYYS: check for mod/broadcaster

pajbot/modules/basic/roll.py
```diff
@@ -0,0 +1,158 @@
...
+from pajbot.modules import BaseModule, ModuleSetting
+from pajbot.modules.basic import BasicCommandsModule
+
+log = logging.getLogger(__name__)
+
+
+class RollModule(BaseModule):
+    ID = __name__.split(".")[-1]
+    NAME = "Roll"
+    DESCRIPTION = "Allows users to roll a random number, which can also be used as timeout times"
```
Felanbird, line 13:
```suggestion
    DESCRIPTION = "Allows users to roll a random number, which can also be used as timeout durations"
```

pajbot/modules/basic/roll.py
ALazyMeme: I was under the impression that pajbot automatically checked for this in the timeout function, but I can't seem to find it anywhere. Will have to talk to pajlada about why that is.

pajlada:
Module bloated, strip it down to do what it's actually meant to do (which is time yourself out a random amount of time).
Don't use word "roll" if it's not in the content of dice.

pajlada (approved):

PR #1725, opened 2022-01-30
Add a `seconds_conversion` function

ALazyMeme:
Pull request checklist:

### DIFF
--- a/pajbot/utils/seconds_conversion.py
+++ b/pajbot/utils/seconds_conversion.py
@@ -1,0 +1,23 @@
+def seconds_conversion(self, value: int, unit: str="second"):
+    unit = unit.lower()
+
+    if unit.endswith(s):
+        unit = unit[:-1]
+
+    if unit == "second":
+        return value
+
+    if unit == "minute":
+        return value * 60
+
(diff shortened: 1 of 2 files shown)


pajbot/utils/seconds_conversion.py
```diff
@@ -0,0 +1,23 @@
+def seconds_conversion(self, value: int, unit: str = "second"):
```
pajlada, line 1: Typing: Add return value

pajbot/utils/seconds_conversion.py
```diff
@@ -0,0 +1,23 @@
...
+        return value * 3600
+
+    if unit == "day":
+        return value * 86400
+
+    if unit == "week":
+        return value * 604800
+
+    if unit == "month":
+        return value * 2629746
```
pajlada, line 23: Missing return value or exception if the unit doesn't match. Potentially store the units in a dict instead with a multiplier?
```
valid_units = {
"second": 1,
"minute": 60,
... etc
}
```

pajlada (changes requested):
Add some tests (see other utils)

PR #1806, opened 2022-04-02
Update menu based on module state without forcing web app restart

ALazyMeme:
Pull request checklist:

I didn't want to have to manually define this, so I've added an option to just omit the modules config entry altogether.


ALazyMeme:
@pajlada I noticed that the web service has to be restarted each time, is there a way for it to update live? Or will it just need a restart every now and then?

pajlada (approved):

PR #1815, opened 2022-04-04
Remove autobuild from codeql

ALazyMeme:
Based on what I'm reading [here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages), autobuilding is something that can be removed for projects that have a different building method. Based on this information, we should be ok to remove it; which should also speed up the codeql process.


ALazyMeme:
Also the fact that prompting to create a new workflow, comments out the autobuild code:

image

ALazyMeme:
Managed to shave off roughly 2 minutes by removing this (somehow) - according to the python test on this pr and the current master build. Could just be a fluke though

RAnders00 (approved):

PR #1844, opened 2022-04-15
Update chart script to 3.7.1

ALazyMeme:
Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 - Bugfix: Fix banphrase API not properly returning matching banphrase. (#1823)
 - Bugfix: Fix toggling of playsound module from the playsound admin page. (#1825)
 - Bugfix: Fix errors in the API not properly returning a JSON response. (#1833)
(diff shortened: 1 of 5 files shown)


ALazyMeme:
I made the change entry intentionally broad - these changes shouldn't really affect users, so I plan on grouping multiple prs into the 1 changelog entry.

ajinkyagitrepo (approved):
why do i have access to this

pajlada:
Have you tested this in the sense of actually having a chart show up? or just "doesn't crash xd"

ALazyMeme:
> Have you tested this in the sense of actually having a chart show up? or just "doesn't crash xd"

I tested the CLR page to make sure it worked (since that's where it seems to be used). I'm not actually quite sure where the chart is though 🤔

RAnders00:
I believe it‘s used when a prediction (via the prediction module) is ongoing. It shows a pie chart iirc

ALazyMeme:
> I believe it‘s used when a prediction (via the prediction module) is ongoing. It shows a pie chart iirc

Looks like it might actually be the hsbet module, might see to removing it as well.

pajlada:
Closing in favour of #1864

Issue #1772, opened 2022-03-09
Reloading VIPs fails if at least one user with non-ascii display name is a VIP


zneix:
See https://github.com/twitchdev/issues/issues/73#issuecomment-1013881657 - basically response of `/vips` which is sent to chat in `pajbot/modules/vip_refresh.py` to determine a list of VIP users, gets a response with display names, which could be problematic, since we all know that display names on Twitch don't always match the login name of the associated user.

Example of a `/vips` response that causes issues:
`@msg-id=vips_success :tmi.twitch.tv NOTICE #zneix :The VIPs of this channel are: ALazyMeme, Amayii, AnatoleAM, CollapsingWave, Eltefan, IceCreamDataBase, Leppunen, NotKarar, PepegePaul, Smaczny, Somso2e, Supinic, TETYYS, TeoTheParty, TranRed, WiFeSeN, YOSEFSA7, YOSEFSAA7, YUNG_RANDD, YungLPR, apa420, boring_nick, cbdg, fabZeef, flex3rs, foUrtf, jammehcow, matthewde, muskatico, pajlada, randers, ravi0li_, rl_grey, romyDank, sbus, talenq, teischEnte, tolekk, whereismymnd, xVeroNy, ツーツーツーツーツー
ツーツーツーツーツ.`

The issue: last displayname is (as of time of writing this issue) is of a user with login name `pilzkman`:

<details>
<summary>Stack trace</summary>

```
Mar 09 20:07:41 shungite pajbot@zneix[2542913]: [2022-03-09 20:07:41,008] [ERROR  ] Logging an uncaught exception (ActionQueue)
Mar 09 20:07:41 shungite pajbot@zneix[2542913]: Traceback (most recent call last):
Mar 09 20:07:41 shungite pajbot@zneix[2542913]:   File "/usr/lib/python3.9/concurrent/futures/thread.py", line 52, in run
Mar 09 20:07:41 shungite pajbot@zneix[2542913]:     result = self.fn(*self.args, **self.kwargs)
Mar 09 20:07:41 shungite pajbot@zneix[2542913]:   File "/opt/pajbot/pajbot/utils/time_method.py", line 22, in wrap
Mar 09 20:07:41 shungite pajbot@zneix[2542913]:     ret = f(*args, **kwargs)
Mar 09 20:07:41 shungite pajbot@zneix[2542913]:   File "/opt/pajbot/pajbot/modules/vip_refresh.py", line 70, in _process_vip_logins
[...]

PR #1862, opened 2022-04-22
Workaround for vip_refresh failure

zneix:
Pull request checklist:

As described throughtly in GH-1772, `vips_success` NOTICE returns display names, which could contain non-ascii characters - these special names make Helix whole helix request fail and that is what I'm trying to avoid here.

Closes: GH-1772

### DIFF
--- a/pajbot/modules/vip_refresh.py
+++ b/pajbot/modules/vip_refresh.py
@@ -27,7 +27,7 @@ class VIPRefreshModule(BaseModule):
     UPDATE_INTERVAL = 10  # minutes
 
     # This regex matches login names but not display names if they contain any non-ascii characters or spaces
-    LOGIN_REGEX = re.compile(r"^[a-z0-9](?:\w{1,24})?$", re.IGNORECASE)
+    LOGIN_REGEX = re.compile(r"^[a-z0-9]\w{0,24}$", re.IGNORECASE)
 
     def __init__(self, bot):
         super().__init__(bot)
--- a/pajbot/modules/vip_refresh.py
+++ b/pajbot/modules/vip_refresh.py
@@ -40,7 +40,7 @@ class VIPRefreshModule(BaseModule):
         self.bot.privmsg("/vips")
 
     @staticmethod
-    def _parse_pubnotice_for_vips(self, msg_id, message):
+    def _parse_pubnotice_for_vips(msg_id, message):
         if msg_id == "no_vips":
             return []
 
@@ -68,7 +68,7 @@ class VIPRefreshModule(BaseModule):
         if channel != self.bot.streamer.login:
             return
 
-        vip_logins = self._parse_pubnotice_for_vips(self, msg_id, message)
+        vip_logins = self._parse_pubnotice_for_vips(msg_id, message)
 
         if vip_logins is not None:
             self.bot.action_queue.submit(self._process_vip_logins, vip_logins)
--- a/pajbot/modules/vip_refresh.py
+++ b/pajbot/modules/vip_refresh.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from pajbot.managers.db import DBManager
 from pajbot.managers.handler import HandlerManager
@@ -26,6 +27,7 @@ class VIPRefreshModule(BaseModule):
     def __init__(self, bot):
         super().__init__(bot)
         self.scheduled_job = None
+        self.username_regex = re.compile("^[a-zA-Z0-9](?:\w{1,24})?$")
 
     def update_vip_cmd(self, bot, source, **rest):
         # TODO if you wanted to improve this: Provide the user with feedback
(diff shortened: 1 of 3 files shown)


pajbot/modules/vip_refresh.py
```diff
@@ -37,14 +39,20 @@ def _update_vips(self):
...
         if msg_id == "vips_success":
             if message.startswith("The VIPs of this channel are: ") and message.endswith("."):
                 message = message[30:-1]  # 30 = length of the above "prefix", -1 removes the dot at the end
-                return message.split(", ")
+                vips = message.split(", ")
+                # Response of "vips_success" returns display names, which could contain non-ascii characters.
+                # For now, we filter these out since these cannot be resolved through helix, see https://github.com/pajbot/pajbot/issues/1772
+                filtered_vips = [vip for vip in vips if re.match(self.username_regex, vip)]
+                if len(vips) != len(filtered_vips):
+                    log.warning("Some non-ascii display names were filtered out from vips_success NOTICE response.")
```
zneix, line 54: I'm not sure about this error message, but I decided I'd give at least some indication of working around this issue. Because, even with this merged in, there will still be some users not acknowledged as VIPs (the ones with non-ascii usernames), so I think it's good to show that something's not right.

pajbot/modules/vip_refresh.py
```diff
@@ -26,6 +27,7 @@ class VIPRefreshModule(BaseModule):
     def __init__(self, bot):
         super().__init__(bot)
         self.scheduled_job = None
+        self.username_regex = re.compile(r"^[a-z0-9](?:\w{1,24})?$", re.IGNORECASE)
```
RAnders00, line 30: I would prefer if this was static and not stored on the instance

pajbot/modules/vip_refresh.py
```diff
@@ -37,14 +39,20 @@ def _update_vips(self):
         self.bot.privmsg("/vips")
 
     @staticmethod
-    def _parse_pubnotice_for_vips(msg_id, message):
+    def _parse_pubnotice_for_vips(self, msg_id, message):
```
RAnders00, line 42: Would then also not require the `self` parameter here

pajbot/modules/vip_refresh.py
```diff
@@ -59,7 +67,7 @@ def _on_pubnotice(self, channel: str, msg_id, message) -> None:
         if channel != self.bot.streamer.login:
             return
 
-        vip_logins = self._parse_pubnotice_for_vips(msg_id, message)
+        vip_logins = self._parse_pubnotice_for_vips(self, msg_id, message)
```
RAnders00, line 70: Same here

RAnders00 (changes requested):

pajbot/modules/vip_refresh.py
RAnders00: Oh, it should also be called LOGIN_REGEX or something like that, since I want to be strict about separating `id`, `login` and `name`

RAnders00: See for example https://github.com/pajbot/pajbot/blob/3610d47181be2b3d376c286e8b91bc52d75574f3/pajbot/models/user.py#L43

RAnders00:
Oh, while you‘re at it, do you think you could add unit tests for this static method? E.g. in https://github.com/pajbot/pajbot/tree/master/pajbot/tests/modules

I‘d like to make sure the method can handle not just korean display names, but also things like names with spaces (`Riot Games`) or those users that have an extra space at the end of their display name (`ExampleUser `)

pajbot/modules/vip_refresh.py
pajlada: I'm fine having this here :+1:

pajbot/modules/vip_refresh.py
pajlada: Seems good :+1:

pajlada:
@RAnders00 I went ahead and renamed it to `LOGIN_REGEX` and made it static.
I also moved the actual filtering to a separate static function to make it easier to test. So the tests that have been added now only test the display-name filtering function, not how we handle a full pubnotice message

pajbot/modules/vip_refresh.py
```diff
@@ -23,6 +26,9 @@ class VIPRefreshModule(BaseModule):
 
     UPDATE_INTERVAL = 10  # minutes
 
+    # This regex matches login names but not display names if they contain any non-ascii characters or spaces
+    LOGIN_REGEX = re.compile(r"^[a-z0-9](?:\w{1,24})?$", re.IGNORECASE)
```
RAnders00, line 30:
```suggestion
    LOGIN_REGEX = re.compile(r"^[a-z0-9]\w{0,24}$", re.IGNORECASE)
```

RAnders00 (changes requested):

RAnders00 (approved):

PR #1873, opened 2022-04-25
Show both display and login name for non-ascii display names

RAnders00:
Pull request checklist:


RAnders00:
Needs testing

ALazyMeme:
I think this would look ok in web, but would look weird in chat messages.

pajlada (commented):
I like the idea of this but not where it's placed, I'd prefer this to be a new property in the `User` model that commands and templates explicitly choose to use. Maybe:
```python3
@property
def name_and_login(self) -> str:
    """
    Return the name and login of the user if the name contains non-ascii characters (e.g. korean hangul characters), and show only the name if they're the same with only casing changes (e.g. Pajlada and pajlada)
    """
    if self.name.lower() == self.login:
        return self.name
    else:
        return f"{self.name} ({self.login})"
```

ALazyMeme:
> I like the idea of this but not where it's placed, I'd prefer this to be a new property in the `User` model that commands and templates explicitly choose to use. Maybe:

Happy with this solution, if @RAnders00 agrees, then I'll close my pr.

RAnders00:
That sounds like a good solution. I‘ll continue work on this PR then.

ALazyMeme:
If someone figures out a better looking changelog entry or title, let me know.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 - Major: Potentially breaking, filters in commands are now applied even if the result of a substitution returns an error. (#1973)  
   This makes all `or_...` filters a lot more useful, but it may mean that some other filters will need some additional error handling.  
   This will require some experimentation, reporting errors in our GitHub issues for this is greatly appreciated.
+- Minor: Added `name_and_login` user property, which displays a user's login name if their display name is not in English. (#1873)
```
RAnders00, line 11:
```suggestion
- Minor: The website now shows the user's login name in addition to the display name if the user's name is not in Latin characters. (#1873)
```

pajbot/models/user.py
```diff
@@ -426,6 +426,16 @@ def find_by_login(db_session: Session, login: str) -> Optional[User]:
     def find_by_id(db_session: Session, id: str) -> Optional[User]:
         return db_session.query(User).filter_by(id=id).one_or_none()
 
+    @property
+    def name_and_login(self) -> str:
```
RAnders00, line 430: I'd at least like to debate the naming of this function, maybe `display_verbose()` and remove the `@property`? So that ` {{ user.name_and_login }}` becomes `{{ user.display_verbose() }}` - I think this would be clearer. Am open to comments and other people's thoughts on this though.

CHANGELOG.md
ALazyMeme: Wouldn't this make it seem like it's replaced in EVERY scenario, when it's actually not?

pajlada: It does it in as many places as makes sense for now - we can fill in the gap if we find a place where it makes sense to do it after this PR

pajbot/models/user.py
pajlada: Ended up with a "happy medium", calling it `name_verbose`, making sure the comment was clear, but keeping it a property so it's consistent with `.name` and `.login` in templates

pajlada (approved):

PR #1874, opened 2022-04-25
Update `me` method in `send_message_to_user` function to include user ping

ALazyMeme:
Pull request checklist:

Currently the me method in the `send_message_to_user` function is completely useless. It doesn't do anything on top of the method in the `send_message` and requires extra arguments. Given the defining part between both functions is sending the message TO THE USER, we should ping the user in the me method (same as the say method).

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -715,7 +715,7 @@ class Bot:
         elif method == "whisper":
             self.whisper(user, message)
         elif method == "me":
-            self.me(message)
+            self.me(f"@{user.name}, {message}")
         elif method == "reply":
             if event.type in ["action", "pubmsg"]:
                 msg_id = next(tag["value"] for tag in event.tags if tag["key"] == "id")


ALazyMeme:
Given this will work pretty much exactly the same as the `say` method (which already works), I don't think this needs testing.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 - Bugfix: Fix errors in the API not properly returning a JSON response. (#1833)
 - Bugfix: Fix command "Check message" option not being modifiable. (#1845)
 - Bugfix: Work around no VIPs being refreshed through VIP refresh module in some cases. (#1862)
+- Dev: Updated `me` method in `send_message_to_user` function to include user ping. (#1874)
```
RAnders00, line 22: Should be minor

RAnders00 (changes requested):

PR #1875, opened 2022-04-25
Remove `safe_say` function in favour of the `send_message` function

ALazyMeme:
Pull request checklist:

Relies on https://github.com/pajbot/pajbot/pull/1874 being merged.


ALazyMeme:
I can confirm trivia on this pr does not work, but I don't see any errors on my console. It just tells you trivia has started, then never posts any messages:
image

ALazyMeme:
@pajlada any chance you could take a look? I tried looking into the functions and couldn't find the reason for these messages not posting.

pajlada:
> @pajlada any chance you could take a look? I tried looking into the functions and couldn't find the reason for these messages not posting.

I don't want to spend time on something that works before this change, but doesn't work after this change. Once everything is confirmed functional you can request my review

PR #1893, opened 2022-04-30
Fix web command list

### LINKED ISSUE #1891: Clicking on commands does nothing on commands page
Clicked on a command, it doesnt do anything

image

Seems to be a change that occurred between ffdacd45 and dabd1f1d (since zneix's bot works just fine on ffdacd45)

pajlada:
- Use `rjsmin` instead of `jsmin`
- Add changelog entry

Fixes #1891

Pull request checklist:


ALazyMeme (approved):

RAnders00:
@pajlada Is the behaviour broken in the last release? Or was pajbot never released with the broken behaviour?
(Just asking because in the latter case a changelog entry shouldn‘t be added.)

pajlada:
> the

It was never in a release, can amend it in the release pr

RAnders00:
> the

PR #1918, opened 2022-05-08
Convert `MessageAction.get_argument_value` to a global function

pajlada:
Allow `get_argument_value` to accept an optional index, making it easier
to use from `Substitution` which doesn't always offer an index

`get_argument_value` now does the index-shifting (1=0) itself, instead
of relying on the user to do it for them. This allows the `index`
argument to be the same as the user writes in commands.

Pull request checklist:


pajlada:
Looks good to me! :+1:

PR #1973, opened 2022-06-04
Filters in commands are now applied even if the result of a substitution returns an error

### LINKED ISSUE #1966: user;1:username|or_broadcaster doesn't work as expected
If I make a command with this variable $(user;1:username|or_broadcaster) I would expect it to find the user in the first argument, and if that fails it should return the broadcasters username.
But currently it doesn't return a message at all, if no user is found. Only works if an argument is given.

Example:
!add command test https://tmi.twitch.tv/group/user/$(user;1:username|or_broadcaster)/chatters

!test forsen - Returns: https://tmi.twitch.tv/group/user/forsen/chatters

!test - Returns nothing

pajlada:
- Filters are now applied even if the result of a substitution is None
- Add changelog entry

Fixes #1966

This makes all `or_...` filters a lot more useful, but it may mean that
some other filters will need some additional error handling.

I think the tradeoff is worth it, and that we can take care of errors
with other filters as they come up.

How to test:
`!edit command foof a $(user;1:username|or_broadcaster) b`
Before fix:
`!foof` - no reply
`!foof forsen` - reply "a forsen b"

After fix:
`!foof` - reply "a STREAMER b"
`!foof forsen` - reply "a forsen b"

Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
 - Breaking: Removed fatoverlay and crazyoverlay. These were alternatives to `/clr/overlay/<number>`, e.g. `/clr/fatoverlay/<number>`. If you don't know what these are, or if you never used these, then this will not affect you. (#1946)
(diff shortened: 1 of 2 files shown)


ALazyMeme (approved):
forsen

PR #2064, opened 2022-08-06
Add Wide emote limit module

Foretack:
Pull request checklist:

Adds a new module that allows the limiting of wide 7TV & FFZ emotes.

image

### DIFF
--- a/pajbot/managers/emote.py
+++ b/pajbot/managers/emote.py
@@ -220,6 +220,9 @@ class EmoteManager:
                 "2": EmoteManager.twitch_emote_url(emote_id, "2.0"),
                 "4": EmoteManager.twitch_emote_url(emote_id, "3.0"),
             },
+            # Twitch emotes are always size-fixed
+            max_width=112,
+            max_height=112
         )
 
     @staticmethod
--- a/pajbot/managers/emote.py
+++ b/pajbot/managers/emote.py
@@ -220,7 +220,7 @@ class EmoteManager:
                 "2": EmoteManager.twitch_emote_url(emote_id, "2.0"),
                 "4": EmoteManager.twitch_emote_url(emote_id, "3.0"),
             },
-            # Twitch doesn't provide the dimensions of emotes. However, subscriber emotes are always size-fixed
+            # Twitch gives no data regarding this, but it can be assumed that this will be the emote size most of the time 
             max_width=112,
             max_height=112
         )
--- a/pajbot/managers/emote.py
+++ b/pajbot/managers/emote.py
@@ -220,9 +220,9 @@ class EmoteManager:
                 "2": EmoteManager.twitch_emote_url(emote_id, "2.0"),
                 "4": EmoteManager.twitch_emote_url(emote_id, "3.0"),
             },
-            # Twitch gives no data regarding this, but it can be assumed that this will be the emote size most of the time 
+            # Twitch gives no data regarding this, but it can be assumed that this will be the emote size most of the time
             max_width=112,
-            max_height=112
+            max_height=112,
         )
 
     @staticmethod
--- a/pajbot/managers/emote.py
+++ b/pajbot/managers/emote.py
@@ -220,7 +220,7 @@ class EmoteManager:
                 "2": EmoteManager.twitch_emote_url(emote_id, "2.0"),
                 "4": EmoteManager.twitch_emote_url(emote_id, "3.0"),
             },
-            # Twitch emotes are always size-fixed
+            # Twitch doesn't provide the dimensions of emotes. However, subscriber emotes are always size-fixed
             max_width=112,
             max_height=112
         )
--- a/pajbot/models/emote.py
+++ b/pajbot/models/emote.py
@@ -13,12 +13,12 @@ class Emote:
     :type id: str
     :var urls: Dict mapping size (e.g. "1", "2", or "4") to an URL depicting the emote
     :type urls: dict[str, str]
-    :ivar max_width: The maximum width the emote can have.
+    :ivar max_width: Width of the largest variant of the emote.
     :type max_width: int
-    :ivar max_height: The maximum height the emote can have.
+    :ivar max_height: Height of the largest variant of the emote.
     :type max_height: int"""
 
-    def __init__(self, code: str, provider: str, id: str, urls: Dict[str, str], max_width: int = 128, max_height: int = 128) -> None:
+    def __init__(self, code: str, provider: str, id: str, urls: Dict[str, str], max_width: int = 112, max_height: int = 112) -> None:
         self.code = code
         self.provider = provider
         if not isinstance(id, str):
--- a/pajbot/models/emote.py
+++ b/pajbot/models/emote.py
@@ -11,16 +11,22 @@ class Emote:
     :type provider: str
     :ivar id: Provider-specific ID, e.g. the emote ID for twitch, or the has for bttv.
     :type id: str
+    :ivar max_width: The maximum width the emote can have.
+    :type max_width: int
     :var urls: Dict mapping size (e.g. "1", "2", or "4") to an URL depicting the emote
-    :type urls: dict[str, str]"""
+    :type urls: dict[str, str]
+    :ivar max_height: The maximum height the emote can have.
+    :type max_height: int"""
 
-    def __init__(self, code: str, provider: str, id: str, urls: Dict[str, str]) -> None:
+    def __init__(self, code: str, provider: str, id: str, urls: Dict[str, str], max_width: int, max_height: int) -> None:
         self.code = code
         self.provider = provider
         if not isinstance(id, str):
             raise ValueError("id parameter must be a string")
         self.id = id
         self.urls = urls
+        self.max_width = max_width
+        self.max_height = max_height
 
     def __eq__(self, other) -> bool:
         if not isinstance(other, Emote):
@@ -35,7 +41,7 @@ class Emote:
         return f"[{self.provider}] {self.code}"
 
     def jsonify(self) -> Dict[str, Any]:
-        return {"code": self.code, "provider": self.provider, "id": self.id, "urls": self.urls}
+        return {"code": self.code, "provider": self.provider, "id": self.id, "urls": self.urls, "max_width": self.max_width, "max_height": self.max_height}
 
     @staticmethod
     def from_json(json_data) -> Emote:
--- a/pajbot/models/emote.py
+++ b/pajbot/models/emote.py
@@ -18,7 +18,9 @@ class Emote:
     :ivar max_height: Height of the largest variant of the emote.
     :type max_height: int"""
 
-    def __init__(self, code: str, provider: str, id: str, urls: Dict[str, str], max_width: int, max_height: int) -> None:
+    def __init__(
+        self, code: str, provider: str, id: str, urls: Dict[str, str], max_width: int, max_height: int
+    ) -> None:
         self.code = code
         self.provider = provider
         if not isinstance(id, str):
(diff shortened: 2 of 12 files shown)


pajbot/managers/emote.py
```diff
@@ -220,6 +220,9 @@ def twitch_emote(emote_id: str, code: str) -> Emote:
                 "2": EmoteManager.twitch_emote_url(emote_id, "2.0"),
                 "4": EmoteManager.twitch_emote_url(emote_id, "3.0"),
             },
+            # Twitch emotes are always size-fixed
```
pajlada, line 223: Technically not true

pajbot/models/emote.py
```diff
@@ -12,15 +12,21 @@ class Emote:
     :ivar id: Provider-specific ID, e.g. the emote ID for twitch, or the has for bttv.
     :type id: str
     :var urls: Dict mapping size (e.g. "1", "2", or "4") to an URL depicting the emote
-    :type urls: dict[str, str]"""
+    :type urls: dict[str, str]
+    :ivar max_width: The maximum width the emote can have.
+    :type max_width: int
+    :ivar max_height: The maximum height the emote can have.
+    :type max_height: int"""
```
pajlada, line 19: The maximum N of the largest scale emote? something like that is probably clearer

pajbot/models/emote.py
```diff
@@ -12,15 +12,21 @@ class Emote:
...
 
-    def __init__(self, code: str, provider: str, id: str, urls: Dict[str, str]) -> None:
+    def __init__(self, code: str, provider: str, id: str, urls: Dict[str, str], max_width: int = 128, max_height: int = 128) -> None:
```
pajlada, line 21: You use 128 here and 112 in other places where you define a "default" - is a default value here necessary?

pajbot/modules/__init__.py
```diff
@@ -100,6 +101,7 @@
     EightBallModule,
     EmoteComboModule,
     EmoteLimitModule,
+    WideEmoteLimitModule,
```
pajlada, line 104: add your module in alphabetical order

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,121 @@
...
+from pajbot.managers.handler import HandlerManager
+from pajbot.modules import BaseModule, ModuleSetting
+
+log = logging.getLogger(__name__)
+
+
+class WideEmoteLimitModule(BaseModule):
+    ID = __name__.split(".")[-1]
+    NAME = "Wide Emote Limit"
+    DESCRIPTION = "Times out users who post too many wide 7TV or FFZ emotes"
```
pajlada, line 12: Add a `LONG_DESCRIPTION` or whatever the field is called explaining that we only know the width of 7tv and ffz emotes, bttv and twitch emotes don't give us their size

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,121 @@
...
+            label="Maximum number of wide emotes that can be posted",
+            type="number",
+            required=True,
+            placeholder="",
+            default=15,
+            constraints={"min_value": 1, "max_value": 167},
+        ),
+        ModuleSetting(
+            key="max_width",
+            label="Maximum width of emotes in pixels. Emotes exceeding this width will be counted as wide",
```
pajlada, line 26: Key and label can be made clearer

The intent of the setting is to specify a width that makes the bot consider the emote wide
Should also make sure to make it clear if the user specified 128 whether a 128-width emote is wide or a 129-width emote is wide

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,121 @@
...
+        if self.bot.is_online and not self.settings["enable_in_online_chat"]:
+            return True
+
+        if not self.bot.is_online and not self.settings["enable_in_offline_chat"]:
+            return True
+        
+        if self.settings["allow_subs_to_bypass"] and source.subscriber is True:
+            return True
+
+        exceeding_emotes = [i.emote.max_width for i in emote_instances if i.emote.max_width > self.settings["max_width"]]
```
pajlada, line 103: `wide_emotes` would be a better variable name for this

pajlada (changes requested):

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,121 @@
...
+            type="boolean",
+            required=True,
+            default=False,
+        ),
+        ModuleSetting(
+            key="enable_in_online_chat", label="Enabled in online chat", type="boolean", required=True, default=True
+        ),
+        ModuleSetting(
+            key="enable_in_offline_chat", label="Enabled in offline chat", type="boolean", required=True, default=True
+        ),
```
ALazyMeme, line 71: Can we replace `timeout_duration` with an online and offline timeout duration? Some streamers may want a different timeout length in offline chat compared to online chat.

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,121 @@
...
+            constraints={"min_value": 112, "max_value": 1000},
+        ),
+        ModuleSetting(
+            key="bypass_level",
+            label="Level to bypass module",
+            type="number",
+            required=True,
+            placeholder="",
+            default=420,
+            constraints={"min_value": 100, "max_value": 1000},
```
ALazyMeme, line 40: Anything level 500+ is counted as mod, so `max_value` is pointless for anything over that. Probably best to set `max_value` to 499

pajbot/modules/__init__.py
```diff
@@ -37,6 +37,7 @@
 from pajbot.modules.dummy import DummyModule
 from pajbot.modules.eightball import EightBallModule
 from pajbot.modules.emote_limit import EmoteLimitModule
+from pajbot.modules.wide_emote_limit import WideEmoteLimitModule
```
ALazyMeme, line 40: add your module in alphabetical order

pajbot/modules/wide_emote_limit.py
ALazyMeme, line 12:
```suggestion
    DESCRIPTION = "Punishes users that post too many wide 7TV or FFZ emotes."
```

pajbot/models/emote.py
Foretack: I've updated the value to 112 for the sake of consistency, but as to whether I should set a default or not: I encountered issues with the bot trying to load emotes from the cache at the start, since the cached value didn't contain these properties.  So I thought I'd set a default to avoid forcing everyone to clear their emotes cache upon updating

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,131 @@
...
+            type="boolean",
+            required=True,
+            default=False,
+        ),
+        ModuleSetting(
+            key="enable_in_online_chat", label="Enabled in online chat", type="boolean", required=True, default=True
+        ),
+        ModuleSetting(
+            key="enable_in_offline_chat", label="Enabled in offline chat", type="boolean", required=True, default=True
+        ),
```
ALazyMeme, line 81: This isn't needed anymore, since setting `timeout_offline` or `timeout_online` should turn them off respectively.

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,131 @@
...
+        ModuleSetting(
+            key="timeout_reason",
+            label="Timeout Reason",
+            type="text",
+            required=False,
+            placeholder="",
+            default="Your message contains too many wide emotes",
```
ALazyMeme, line 88:
```suggestion
            default="Too many wide emotes",
```

pajbot/modules/wide_emote_limit.py
ALazyMeme: Adding @pajlada for opinion

pajbot/managers/emote.py
mmattbtw: here are some still available wide emotes for testing, if needed: https://emotes.raccatta.cc/twitch/thegdstudio

pajbot/modules/wide_emote_limit.py
pajlada: Sure

pajlada: This is false @ALazyMeme 
image

testaccount_420 or :joy: in this case is level 500 but not a moderator. moderator returns false, so keeping the max at 1000 is perfectly fine

pajbot/models/emote.py
pajlada: In that case, the cache should be cleared by this PR. This should be done using a migration revision - see `pajbot/migration_revisions/redis/0002_fix_bad_quest_emote_json.py`
you'll need to make a new revision `0003_asdasd.py` that deletes all cached emotes from Redis

pajlada (approved):
Works as expected :+1:

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,127 @@
...
+        if self.bot.is_online and self.settings["timeout_online"] == 0:
+            return True
+
+        if not self.bot.is_online and self.settings["timeout_offline"] == 0:
+            return True
+
+        if self.settings["allow_subs_to_bypass"] and source.subscriber is True:
+            return True
+
+        wide_emotes = [
```
RAnders00, line 107: There‘s no need to materialize this filtering into a list. `wide_emotes = ( 1 for i in … if … )` would be preferable. You can then do `sum(wide_emotes)`

RAnders00 (changes requested):

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,125 @@
...
+        if self.bot.is_online and self.settings["timeout_online"] == 0:
+            return True
+
+        if not self.bot.is_online and self.settings["timeout_offline"] == 0:
+            return True
+
+        if self.settings["allow_subs_to_bypass"] and source.subscriber is True:
+            return True
+
+        wide_emotes = [1 for i in emote_instances if i.emote.max_width > self.settings["emote_max_width"]]
```
RAnders00, line 107: You didn't do what I suggested - the main point was the usage of the `()` instead of `[]` - as one is an iterator, the other one constructs a new list. Maybe this stackoverflow thread would help you understand: https://stackoverflow.com/q/53376146/4464702

RAnders00 (changes requested):

pajbot/modules/wide_emote_limit.py
Foretack: I wasn't aware, let me know if I implemented it correctly. Thanks 👍

RAnders00: Yeah, that looks good! Thanks!

RAnders00 (approved):

pajbot/modules/wide_emote_limit.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def on_message(self, source, message, emote_instances, msg_id, **rest):
+        if source.level >= self.settings["bypass_level"] or source.moderator is True:
+            return True
+
+        if self.bot.is_online and self.settings["timeout_online"] == 0:
+            return True
+
+        if not self.bot.is_online and self.settings["timeout_offline"] == 0:
+            return True
```
ALazyMeme, line 102: This returns if `timeout_online` and `timeout_offline is set to 0. Which stops the delete message function from working.

ALazyMeme (approved):

ALazyMeme:
@pajlada @RAnders00 can one of you pls double check the code changed? just want to make sure it's correct before the pr is merged.

pajlada (approved):

PR #2078, opened 2022-08-15
Updated `!check_sub` to `!checksub` to be consistent with `!checkmod`

ALazyMeme:
Pull request checklist:

I did some function migrations too, but those aren't major enough to document.


pajlada:
What purpose does this PR serve? Because this is potentially a breaking change for anyone with a funccommand pointing at this function

ALazyMeme:
> What purpose does this PR serve?

Specifically to fix the ugly looking command that has an _ in it. Is there a better way of doing it?

pajlada:
This change doesn't change any command, it changes a function name. `!checkmod` is registered in pajbot/modules/basic/checkmod.py
`!checksub` or `!check_sub` or w/e is added manually by calling `!add funccommand checksub check_sub` or `!add funccommand foobarbaz check_sub`. You changing the function name just makes any previous additions of the function invalid

PR #2136, opened 2022-09-30
Add VIP tag support

ALazyMeme:
Pull request checklist:

Added support for the new [Twitch VIP IRC](https://dev.twitch.tv/docs/irc/tags#privmsg-tags) tag. It only shows when the user is actually VIP, so we don't need to do anymore checking aside from the tag just being there.

Tested and works just fine. (Edit: testing was possibly a fluke)

image

### DIFF
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -882,7 +882,7 @@ class Bot:
             # once they are a founder they are always be a founder, regardless if they are a sub or not.
             if not source.founder:
                 source.founder = "founder" in badges
-            source.vip = "vip" in badges
+            source.vip = "vip" in (badges or tags)
 
         if not whisper and source.banned:
             self.ban(
--- a/pajbot/bot.py
+++ b/pajbot/bot.py
@@ -882,7 +882,7 @@ class Bot:
             # once they are a founder they are always be a founder, regardless if they are a sub or not.
             if not source.founder:
                 source.founder = "founder" in badges
-            source.vip = "vip" in (badges or tags)
+            source.vip = "vip" in badges or "vip" in tags
 
         if not whisper and source.banned:
             self.ban(


RAnders00 (approved):

pajbot/bot.py
```diff
@@ -882,7 +882,7 @@ def parse_message(self, message, source, event, tags={}, whisper=False) -> bool:
             # once they are a founder they are always be a founder, regardless if they are a sub or not.
             if not source.founder:
                 source.founder = "founder" in badges
-            source.vip = "vip" in badges
+            source.vip = "vip" in (badges or tags)
```
RAnders00, line 885: Is this some kind of fancy python generator syntax? Because otherwise wouldnt this first evaluate `badges or tags`, which is always `True`? Have you tested that with this change people without VIP status correctly get `vip=False` assigned?

RAnders00 (changes requested):

pajbot/bot.py
RAnders00: I would think it would have to be

```suggestion
            source.vip = "vip" in badges or "vip" in tags
```

ALazyMeme: > Is this some kind of fancy python generator syntax? Because otherwise wouldnt this first evaluate `badges or tags`, which is always `True`? Have you tested that with this change people without VIP status correctly get `vip=False` assigned?

Yes, the part I missed in my screenshot was me testing my own status. See below, you're free to test (or pajlada probably will), but it seemingly works just fine:

image

ALazyMeme: based on my very rough testing, it may be a case of python taking the first value in the brackets (in this case, badges):

```py
number = "56"
sentence = "xd"

print("56" in (number or sentence))
```
> True

```py
number = "56"
sentence = "xd"

print("56" in (sentence or number))
```
> False

ALazyMeme: Subsequent question would be: can we use something like `"vip" in [badges,tags]`?

pajlada (approved):

PR #2141, opened 2022-10-03
Migrate announce function to helix

ALazyMeme:
Pull request checklist:

The changelog entry is general and can be added onto. I didn't want to flood the changelog doc with similar changes on many lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Breaking: Migrated IRC commands to relevant Helix endpoints. This will require re-auth using the `bot_login` endpoint in order to utilize the required functions --> Announce.
 - Minor: Migrated LastFM module to the `reply` response type. (#2118, #2128)
 
 ## v1.62
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## Unversioned
 
-Due to Twitch deprecating the use of **most** IRC commands, we are now using the Helix calls. It's no longer possible to use said these IRC commands within pajbot, instead the bot **must** be re-authenticated in order to grant new permissions to utilize these Helix functions: `moderator:manage:announcements`.
-In short: The bot owner must re-authenticate with the `/bot_login` endpoint for the announcement function to work.
-
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
 - Breaking: Migrated IRC commands to relevant Helix endpoints: Announce. (#2141)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ For your bot to work after these changes, the bot owner **must** re-authenticate
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
-- Breaking: Migrated IRC commands to relevant Helix endpoints: Announce. (#2141)
+- Breaking: Migrated Announce from IRC to Helix. (#2141)
 - Minor: Migrated LastFM module to the `reply` response type. (#2118, #2128)
 
 ## v1.62
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
-- Breaking: Migrated IRC commands to relevant Helix endpoints. This will require re-auth using the `bot_login` endpoint in order to utilize the required functions --> Announce.
+- Breaking: Migrated IRC commands to relevant Helix endpoints. This will require re-auth using the `bot_login` endpoint in order to utilize the required functions --> Announce. (#2141)
 - Minor: Migrated LastFM module to the `reply` response type. (#2118, #2128)
 
 ## v1.62
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unversioned
 
 Due to Twitch deprecating the use of **most** IRC commands, we are now using the Helix calls. It's no longer possible to use said these IRC commands within pajbot, instead the bot **must** be re-authenticated in order to grant new permissions to utilize these Helix functions: `moderator:manage:announcements`.
-In short: The bot owner must re-authenticate with the `/bot_login` endpoint for the announcement function to work. 
+In short: The bot owner must re-authenticate with the `/bot_login` endpoint for the announcement function to work.
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## Unversioned
 
+Due to Twitch deprecating the use of **most** IRC commands, we are now using the Helix calls. It's no longer possible to use said these IRC commands within pajbot, instead the bot **must** be re-authenticated in order to grant new permissions to utilize these Helix functions: `moderator:manage:announcements`.
+In short: The bot owner must re-authenticate with the `/bot_login` endpoint for the announcement function to work. 
+
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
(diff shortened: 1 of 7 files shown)


ALazyMeme:
I have tested and this works just fine. However I do get an error, unsure where the issue in the code exists: https://i.alazymeme.com/TallDeadFile.txt

There's also the import formatting error, which I'm unsure how to fix (running the formatter did nothing).

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -602,3 +602,17 @@ def get_channel_emotes(
             expiry=60 * 60,
             force_fetch=force_fetch,
         )
+
+    def send_chat_announcement(self, bot_id: str, body: Dict[str, str], authorization, channel_id: str = None):
+        if not body:
+            log.error("Invalid call to send_chat_announcement, missing query parameter. message must be specified")
```
RAnders00, line 608: Don‘t do this. Either just don‘t do this check at all (I‘m sure Twitch will send an error response anyways), or at least raise an exception. Especially returning `False` is extremely bad practice here.

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -602,3 +602,17 @@ def get_channel_emotes(
...
+            return False
+
+        response = self.post(
+            "/chat/announcements",
+            {"broadcaster_id": channel_id, "moderator_id": bot_id},
+            authorization=authorization,
+            json=body,
+        )
+
+        return response.status_code == 204
```
RAnders00, line 618: Don‘t return `True` for success here. Return nothing (`-> None` in the type declaration). `self.post` already raises an exception in case the API request failed with a bad status code.

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -602,3 +602,17 @@ def get_channel_emotes(
...
+    def send_chat_announcement(self, bot_id: str, body: Dict[str, str], authorization, channel_id: str = None):
+        if not body:
+            log.error("Invalid call to send_chat_announcement, missing query parameter. message must be specified")
+            return False
+
+        response = self.post(
+            "/chat/announcements",
+            {"broadcaster_id": channel_id, "moderator_id": bot_id},
+            authorization=authorization,
+            json=body,
```
RAnders00, line 615:
```suggestion
            json={"message": message},
```

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -602,3 +602,17 @@ def get_channel_emotes(
             expiry=60 * 60,
             force_fetch=force_fetch,
         )
+
+    def send_chat_announcement(self, bot_id: str, body: Dict[str, str], authorization, channel_id: str = None):
```
RAnders00, line 606:
```suggestion
    def send_chat_announcement(self, sender_id: str, message: str, authorization, channel_id: str):
```
Please also declare a return type

pajbot/bot.py
```diff
@@ -851,17 +852,35 @@ def safe_say(self, message: str, channel: Optional[str] = None) -> None:
...
+        if channel_id is None:
+            channel_id = self.streamer.id
+
         message = utils.clean_up_message(message)
 
-        self.privmsg("/announce " + message[: CHARACTER_LIMIT - 10], channel=channel)
+        try:
+            self.twitch_helix_api.send_chat_announcement(
+                self.bot_user.id,
+                {"message": message[: CHARACTER_LIMIT - 10]},
```
RAnders00, line 871:
```suggestion
                message[: CHARACTER_LIMIT - 10],
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Breaking: Migrated IRC commands to relevant Helix endpoints. This will require re-auth using the `bot_login` endpoint in order to utilize the required functions --> Announce. (#2141)
```
RAnders00, line 7: This needs a more thorough instruction for what the bot operator needs to do

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -602,3 +602,17 @@ def get_channel_emotes(
...
+        if not body:
+            log.error("Invalid call to send_chat_announcement, missing query parameter. message must be specified")
+            return False
+
+        response = self.post(
+            "/chat/announcements",
+            {"broadcaster_id": channel_id, "moderator_id": bot_id},
```
RAnders00, line 613:
```suggestion
            {"broadcaster_id": channel_id, "moderator_id": sender_id},
```

pajbot/apiwrappers/twitch/helix.py
RAnders00, line 606: It would be nice to have a docstring here for the method and its parameters. Documentation should state things like: an exeption is raised if something is not successful, `sender_id` must be the same user as the user authenticated by the given `authorization`, and that the message should (must? Please test the APIs behaviour on this) be cropped before handing it to this function

pajbot/apiwrappers/twitch/helix.py
RAnders00, line 618:
```suggestion
```

RAnders00 (changes requested):
This is just my pleliminary feedback, I‘ll have to look into what causes the error tomorrow (I will set myself a reminder)

pajbot/apiwrappers/twitch/helix.py
ALazyMeme: This was a direct copy from the `modify_channel_information` function. Given we already check this in the `announce` function, this is likely not something that will be used. In saying that, I don't think we should just leave it to Twitch to respond to no message - it's a waste of an API call, especially when the check is already performed in the current version of the `announce` function.

pajbot/apiwrappers/twitch/helix.py
ALazyMeme: I'm not sure if I agree with this change, I intentionally called it `bot_id` to make it as simple and straightforward as possible.

pajbot/apiwrappers/twitch/helix.py
ALazyMeme: `color` can also be in the body, so I'd like to leave that open for addition if anyone wishes to.

pajbot/bot.py
ALazyMeme: `color` can also be in the body, so I'd like to leave that open for addition if anyone wishes to. Either that or I'll end up adding it as an optional field.

pajbot/apiwrappers/twitch/helix.py
ALazyMeme: I ended up adding it.

pajbot/bot.py
ALazyMeme: I ended up adding it.

CHANGELOG.md
ALazyMeme: I have copied something similar from the v1.48 release. Have a gander and let me know what you think.

pajbot/apiwrappers/twitch/helix.py
RAnders00: The `modify_channel_information` needs to be rewritten in that case. Very poor API design that's most certainly not present in all the other apiwrapper functions in here or any of the other apiwrappers.

pajbot/apiwrappers/twitch/helix.py
RAnders00: In that case, don`t do `response = `, just call `self.post()` without using the returned value.

pajbot/bot.py
```diff
@@ -851,17 +852,35 @@ def safe_say(self, message: str, channel: Optional[str] = None) -> None:
...
+            channel_id = self.streamer.id
+
         message = utils.clean_up_message(message)
 
-        self.privmsg("/announce " + message[: CHARACTER_LIMIT - 10], channel=channel)
+        try:
+            self.twitch_helix_api.send_chat_announcement(
+                channel_id,
+                self.bot_user.id,
+                {"message": message[: CHARACTER_LIMIT - 10], "color": color},
```
RAnders00, line 872: I'd still like to not have to construct the POST body here, this is a detail that must be abstracted away by the apiwrapper (that's what the apiwrapper is for)

ALazyMeme:
I can't help but think I should re-add channel name support, call the api, get the id and show a deprecation message in order to not break people's existing functions that call the announce function. Any thoughts?

pajlada (changes requested):

pajlada (approved):

RAnders00 (approved):

PR #2174, opened 2022-10-09
Add `_update_chat_settings` Helix API Function

ALazyMeme:
Pull request checklist:

This PR just adds the base function required to migrate the chat settings over to Helix.


ALazyMeme:
Emote only:
image

Follower only:
image

Subscriber Mode:
image

ALazyMeme:
Note: endpoints requiring a time to be stated, must seemingly also set the initial variable to True at the same time also. eg: `follower_mode=True` must be added in order to use `follower_mode_duration=5`. If you simply pass `follower_mode_duration=5`, it will not work.

pajlada (approved):

RAnders00 (approved):

RAnders00:
I'd like to make a case for omitting the changelog entry: the changelog is for user-facing changes, in my opinion.

PR #2175, opened 2022-10-09
Migrate default chat state module `.followers` to use Helix

ALazyMeme:
Pull request checklist:


ALazyMeme:
Twitch doesn't currently have any error codes documented for the `update_chat_settings` endpoint. I've given it a rough test using insomnia and note that the unauthorized error exists. I've implemented that and catch all other errors in a general exception check. Not much more I can do without the docs.

ALazyMeme:
I managed to check the helix function itself (which works just fine), but aren't currently able to test the function of the module itself. I might be able to further test if I go live tomorrow.
image

ALazyMeme:
I also slightly refactored the code to integrate the `follower_mode_duration` checking into the api request itself as opposed to having a separate if function for it.

pajbot/modules/default_chat_states.py
```diff
@@ -92,6 +94,10 @@ def on_stream_start(self, **rest) -> bool:
             log.warning("on_stream_start failed in DefaultChatStatesModule because bot is None")
             return True
 
+        channel_id = self.bot.streamer.id
+        bot_id = self.bot.bot_user.id
+        authorization = self.bot.bot_token_manager
+
```
RAnders00, line 100: I think I'd prefer these variables be inlined (so use `self.bot.streamer.id` instead of `channel_id` in the `update_follower_mode()` call below)

pajbot/modules/default_chat_states.py
```diff
@@ -117,6 +135,10 @@ def on_stream_stop(self, **rest) -> bool:
             log.warning("on_stream_stop failed in DefaultChatStatesModule because bot is None")
             return True
 
+        channel_id = self.bot.streamer.id
+        bot_id = self.bot.bot_user.id
+        authorization = self.bot.bot_token_manager
+
```
RAnders00, line 141: Same thing here

RAnders00 (dismissed):
Just one small nitpick but apart from that this looks very good. Thank you very much!

ALazyMeme:
Actually there's a bit more to it than we've said here. I've been discussing with pajlada as there's actually more breaking behaviour in the module. Currently we support any format of time, but the API only accepts minutes. So I need to make a conversion function to convert the existing time into minutes.

pajbot/modules/default_chat_states.py
ALazyMeme: The main reason I made these into variables, is that we will need to use these in every single other function in the module. Otherwise we'll be repeating the exact same 3 variables in every function.

pajbot/modules/default_chat_states.py
ALazyMeme: The main reason I made these into variables, is that we will need to use these in every single other function in the module. Otherwise we'll be repeating the exact same self.bot... variables in every function.

pajbot/modules/default_chat_states.py
pajlada: I agree with Randers' take - as we don't make any modifications to the variable, I don't think this is any more or less clean, so just use `self.bot....` variables inline

pajlada (approved):

pajlada:
I tested this out, works as expected with :+1:

PR #2203, opened 2022-10-27
Increased efficiency and speed of subscriber status refresh

ALazyMeme:
Pull request checklist:


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 - Breaking: Migrated Announce from IRC to Helix. (#2141)
 - Breaking: Migrated Delete moderation action from IRC to Helix. (#2173)
 - Minor: Migrated LastFM module to the `reply` response type. (#2118, #2128)
+- Dev: Improved Helix subscriber fetching to utilize the existing data provided by the Helix response. (#2203)
```
RAnders00, line 18:
```suggestion
- Minor: Increased efficiency and speed of subscriber status refresh. (#2203)
```

RAnders00 (changes requested):
Looks good otherwise as far as I can tell! Thank you!

RAnders00 (approved):

pajlada:
Running into this issue where the streamer can appear twice in the endpoint, not sure why this would happen since we should do a `list(set(...))`, needs some more investigation

```
Logging an uncaught exception (ActionQueue)
Traceback (most recent call last):
  File "/home/pajlada/git/pajbot/venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1880, in _execute_context
    self.dialect.do_executemany(
  File "/home/pajlada/git/pajbot/venv/lib/python3.10/site-packages/sqlalchemy/dialects/postgresql/psycopg2.py", line 1000, in do_executemany
    cursor.executemany(statement, parameters)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "subscribers_pkey"
DETAIL:  Key (id)=(11148817) already exists.
```

pajlada (changes requested):
^

RAnders00:
I see the issue. I *assumed* we had this already because I assumed an equality comparison on two `UserBasics` objects would have already been used somewhere else in the code base, but we don‘t have equality and hash implemented for `UserBasics`.

https://github.com/pajbot/pajbot/blob/6b7681f29ce0f4bdb82d22bf44863c091c312094/pajbot/models/user.py#L42
https://stackoverflow.com/a/2909119

As a result of the hash function not being implemented, `set()` is not able to eliminate duplicates.

RAnders00:
Also, I‘d like to argue that the fetcher function could very well return sets already. It‘s quite wasteful to turn list into set and back into list. If we directly used sets all the way (in the `_fetch_all_pages` method and in the `_fetch_X_page` method) we wouldn‘t need to do this. A conversion back to list is probably never necessary (after all, there‘s no inherent order to the list of subs.)

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -340,19 +340,21 @@ def _fetch_subscribers_page(self, broadcaster_id: str, authorization, after_pagi
...
         """Fetch a list of all subscribers (user IDs) of a broadcaster."""
-        subscriber_ids = self._fetch_all_pages(self._fetch_subscribers_page, broadcaster_id, authorization)
+        subscribers = self._fetch_all_pages(self._fetch_subscribers_page, broadcaster_id, authorization)
 
         # Dedupe the list of subscribers since the API can return the same IDs multiple times
-        subscriber_ids = list(set(subscriber_ids))
+        subscribers = list(set(subscribers))
 
-        return subscriber_ids
+        return subscribers
```
RAnders00, line 357:
```suggestion
        return set(subscribers)
```

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -340,19 +340,21 @@ def _fetch_subscribers_page(self, broadcaster_id: str, authorization, after_pagi
...
         return subscribers, pagination_cursor
 
     def fetch_all_subscribers(self, broadcaster_id, authorization):
         """Fetch a list of all subscribers (user IDs) of a broadcaster."""
-        subscriber_ids = self._fetch_all_pages(self._fetch_subscribers_page, broadcaster_id, authorization)
+        subscribers = self._fetch_all_pages(self._fetch_subscribers_page, broadcaster_id, authorization)
 
         # Dedupe the list of subscribers since the API can return the same IDs multiple times
-        subscriber_ids = list(set(subscriber_ids))
+        subscribers = list(set(subscribers))
```
RAnders00, line 355:
```suggestion
```

RAnders00 (approved):

RAnders00:
Oops, I didn‘t mean to select „Approve“, consider it a „Changes requested“, please.

pajlada (approved):
Works as expected

Old: `SubscriberFetchModule::_update_subscribers function took 4342.029 ms`
New: `SubscriberFetchModule::_update_subscribers function took 1413.810 ms`

:+1:

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -347,7 +347,7 @@ def _fetch_subscribers_page(self, broadcaster_id: str, authorization, after_pagi
 
         return subscribers, pagination_cursor
 
-    def fetch_all_subscribers(self, broadcaster_id, authorization):
+    def fetch_all_subscribers(self, broadcaster_id: str, authorization) -> Set[UserBasics]:
```
RAnders00, line 352: The function does not actually return a set!!

PR #2213, opened 2022-11-05
Migrate bans from IRC to Helix

ALazyMeme:
Pull request checklist:

In an effort to maintain compatibility with existing functions, I have:

- Added a `ban_id` function
- Updated `ban` function to utilize `ban_id`
- Updated `ban_login` to get the `userid` of the user being banned

Despite previous conversations, I have chosen NOT to migrate the moderation actions to user IDs yet; else I may break other functions in the process. I think we should work on migrating the functions first to helix, then migrate to user IDs after all the moderation endpoints have been migrated.

Note that I also kept the existing `ban_login` function so we don't break any existing functions that people might have.

### DIFF
--- a/pajbot/apiwrappers/twitch/helix.py
+++ b/pajbot/apiwrappers/twitch/helix.py
@@ -758,3 +758,26 @@ class TwitchHelixAPI(BaseTwitchAPI):
         vips = self._fetch_all_pages(self._fetch_vips_page, broadcaster_id, authorization)
 
         return set(vips)
+
+    def _ban_user(
+        self,
+        broadcaster_id: str,
+        bot_id: str,
+        authorization,
+        duration: Optional[int] = None,
+        reason: str = None,
+        user_id: str = None,
+    ) -> Tuple[str, Optional[str]]:
+        """Calls the Ban User Helix endpoint using the broadcaster_id, bot_id, reason & user_id parameters.
+        broadcaster_id, bot_id, reason & user_id are all required parameters. bot_id must match the user_id in authorization."""
+        response = self.post(
+            "/moderation/bans",
+            {"broadcaster_id": broadcaster_id, "moderator_id": bot_id},
+            authorization=authorization,
+            json={"duration": duration, "reason": reason, "user_id": user_id},
+        )
+
+        created_at = response["data"][0]["created_at"]
+        end_time = response["data"][0]["end_time"]
+
+        return created_at, end_time
--- a/pajbot/apiwrappers/twitch/helix.py
+++ b/pajbot/apiwrappers/twitch/helix.py
@@ -781,3 +781,10 @@ class TwitchHelixAPI(BaseTwitchAPI):
         end_time = response["data"][0]["end_time"]
 
         return created_at, end_time
+
+    def ban_user(self, broadcaster_id: str, bot_id: str, authorization, reason: str, user_id: str) -> str:
+        """Calls the _ban_user function using the reason & user_id parameters.
+        All fields are required. bot_id must match the user_id in authorization."""
+        response = self._ban_user(broadcaster_id, bot_id, authorization, reason, user_id)
+
+        return response[0]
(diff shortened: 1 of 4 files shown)


pajbot/apiwrappers/twitch/helix.py
```diff
@@ -758,3 +758,33 @@ def fetch_all_vips(self, broadcaster_id: str, authorization) -> Set[UserBasics]:
...
         return set(vips)
+
+    def _ban_user(
+        self,
+        broadcaster_id: str,
+        bot_id: str,
+        authorization,
+        duration: Optional[int] = None,
+        reason: str = None,
+        user_id: str = None,
```
pajlada, line 769: `user_id` should be a required parameter

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -758,3 +758,33 @@ def fetch_all_vips(self, broadcaster_id: str, authorization) -> Set[UserBasics]:
...
 
         return set(vips)
+
+    def _ban_user(
+        self,
+        broadcaster_id: str,
+        bot_id: str,
+        authorization,
+        duration: Optional[int] = None,
+        reason: str = None,
```
RAnders00, line 768: shouldn't it be `Optional[str]` here?

RAnders00 (dismissed):

pajbot/apiwrappers/twitch/helix.py
```diff
@@ -758,3 +758,33 @@ def fetch_all_vips(self, broadcaster_id: str, authorization) -> Set[UserBasics]:
         vips = self._fetch_all_pages(self._fetch_vips_page, broadcaster_id, authorization)
 
         return set(vips)
+
+    def _ban_user(
```
pajlada, line 762: The parameters of this function are not easy to understand.

While it strays from some of the norms set up in this class, I think there should be two 'private' functions for using the `Ban User` Helix endpoint:
 - `_ban_user` which omits the `duration` parameter entirely (both in its function definition and in the final API call)
 - `_timeout_user` which requires the `duration` parameter, and forwards it (as expected) to the API call

ALazyMeme: What is the point of having them as internal functions at all? Since they replace the original idea; aren't we better off just making them "public" functions?

If you still suggest that we make them internal functions, do you suggest I also make 2 "public" functions? `ban_user` and `timeout_user`?

pajlada: I'm fine with just making two public functions: `ban_user` and `timeout_user`

pajbot/bot.py
```diff
@@ -638,21 +638,28 @@ def _has_moderation_actions(self) -> bool:
             return False
         return self.thread_locals.moderation_actions is not None
 
-    def _ban(self, login: str, reason: Optional[str] = None) -> None:
-        message = f"/ban {login}"
-        if reason is not None:
-            message += f" {reason}"
-        self.privmsg(message)
+    def _ban(self, userid: str, reason: Optional[str] = None) -> None:
```
pajlada, line 641: Rename `userid` to `user_id`

pajbot/bot.py
```diff
@@ -638,21 +638,28 @@ def _has_moderation_actions(self) -> bool:
...
+            if e.response.status_code == 401:
+                log.error(f"Failed to ban user with id {userid}, unauthorized: {e} - {e.response.text}")
+            else:
+                log.error(f"Failed to ban user with id {userid}: {e} - {e.response.text}")
 
     def ban(self, user: User, reason: Optional[str] = None) -> None:
-        self.ban_login(user.login, reason)
+        self.ban_id(user.id, reason)
+
+    def ban_id(self, userid: str, reason=None) -> None:
```
pajlada, line 653: Rename `userid` to `user_id`

pajbot/bot.py
```diff
@@ -638,21 +638,28 @@ def _has_moderation_actions(self) -> bool:
...
+        self._ban(userid, reason)
 
     def ban_login(self, login: str, reason=None) -> None:
+        userid = self.twitch_helix_api.get_user_id(login)
```
pajlada, line 657: Rename `userid` to `user_id`

pajlada (approved):
Code looks fine to me, feel free to merge in once you've tested it

ALazyMeme:
This works just fine
image

PR #2268, opened 2023-01-03
Use 7TV's new V3 REST API

Nerixyz:
Pull request checklist:

Since the V2 API is considered legacy, this PR updates the 7TV API wrapper to use the new V3 REST API.

I'm not sure what to put in the changelog entry.

### DIFF
--- a/pajbot/apiwrappers/seventv.py
+++ b/pajbot/apiwrappers/seventv.py
@@ -1,10 +1,10 @@
+from typing import Tuple
+
 from pajbot.apiwrappers.base import BaseAPI
 from pajbot.apiwrappers.response_cache import ListSerializer
 from pajbot.models.emote import Emote
 
 from requests import HTTPError
-
-from typing import Tuple
 
 
 class SevenTVAPI(BaseAPI):
@@ -47,7 +47,7 @@ class SevenTVAPI(BaseAPI):
 
     def fetch_global_emotes(self):
         """Returns a list of global 7TV emotes in the standard Emote format."""
-        response = self.get(f"emote-sets/global")
+        response = self.get("emote-sets/global")
 
         return self.parse_emotes(response["emotes"])
 
--- a/pajbot/apiwrappers/seventv.py
+++ b/pajbot/apiwrappers/seventv.py
@@ -29,7 +29,7 @@ class SevenTVAPI(BaseAPI):
                         max_height = file["height"]
                     emote_size += 1
 
-            if len(urls) is 0:
+            if len(urls) == 0:
                 raise ValueError("No file in WEBP format for this emote")
             return (urls, max_width, max_height)
 
--- a/pajbot/apiwrappers/seventv.py
+++ b/pajbot/apiwrappers/seventv.py
@@ -29,7 +29,7 @@ class SevenTVAPI(BaseAPI):
                         max_height = file["height"]
                     emote_size += 1
 
-            if len(urls) == 0:
+            if len(urls) == 0 or max_width == 0 or max_height == 0:
                 raise ValueError("No file in WEBP format for this emote")
             return (urls, max_width, max_height)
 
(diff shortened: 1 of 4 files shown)


ALazyMeme:
> I'm not sure what to put in the changelog entry.

idk I guess if it's not a user noticeable change, just make it a dev entry? or something?

pajbot/apiwrappers/seventv.py
```diff
@@ -2,72 +2,54 @@
...
+            for file in host["files"]:
+                if file["format"] == "WEBP":
+                    name = file["name"]
+                    urls[str(emote_size)] = f"{base_url}/{name}"
+                    if file["width"] > max_width:
+                        max_width = file["width"]
+                    if file["height"] > max_height:
+                        max_height = file["height"]
+                    emote_size += 1
+            return (urls, max_width, max_height)
```
RAnders00, line 31: I think it would be a good idea to raise an exception here (instead of using bogus placeholder data) if none of the files were in the webp format

pajbot/apiwrappers/seventv.py
```diff
@@ -2,72 +2,54 @@
...
 
         emotes = []
-        for emote_data in api_response_data:
-            emote_id = emote_data["id"]
+        for active_emote in api_response_data:
```
RAnders00, line 34: I think I would prefer to keep it at `emote_data`, I've personally never seen somebody refer to the current value in a loop as "active", so naming it that way will just confuse a potential reader

pajbot/apiwrappers/seventv.py
```diff
@@ -78,38 +60,23 @@ def get_global_emotes(self, force_fetch=False):
...
-            return []
+            raise e
 
-        return self.parse_emotes(response["data"]["user"]["emotes"])
+        return self.parse_emotes(response["emote_set"]["emotes"])
 
-    def get_channel_emotes(self, channel_name, force_fetch=False):
+    def get_channel_emotes(self, channel_id, force_fetch=False):
         return self.cache.cache_fetch_fn(
-            redis_key=f"api:7tv:channel-emotes:{channel_name}",
```
RAnders00, line 111: While extremely unlikely, it's possible for a collision to occur between old and new data schemas here.
I think that's an acceptable risk/imperfection though, you basically have to manufacture that setup

I just wanted to point it out

RAnders00 (changes requested):

pajbot/apiwrappers/seventv.py
Nerixyz: 7TV refers to emotes in emote-sets as [active emotes](https://github.com/SevenTV/API/blob/e5992c29bb97118bafad1dbe9f4e752aea34d4ea/data/model/emote-set.model.go#L45-L53). The reasoning is that an active emote can have a different name and/or different flags than the base emote (`emote_data`).

RAnders00: Ah, I didn't know that. All good then!

RAnders00 (approved):

pajbot/apiwrappers/seventv.py
```diff
@@ -1,73 +1,58 @@
...
+                    name = file["name"]
+                    urls[str(emote_size)] = f"{base_url}/{name}"
+                    if file["width"] > max_width:
+                        max_width = file["width"]
+                    if file["height"] > max_height:
+                        max_height = file["height"]
+                    emote_size += 1
+
+            if len(urls) == 0:
+                raise ValueError("No file in WEBP format for this emote")
```
pajlada, line 33: Further exceptions for the max_width or max_height values being zero would be nice

pajlada (changes requested):

pajbot/apiwrappers/seventv.py
Nerixyz: Done in e3da46d.

pajlada (approved):

PR #2319, opened 2023-02-12
vip & mod refreshing can fail if banned/deleted users are in the list

pajlada:
This will ensure we skip any user we get that isn't fully hydrated
Pull request checklist:


pajlada:
image

PR #2320, opened 2023-02-12
Add setting to control how frequently user ranks should be refreshed

pajlada:
Under the `main` section, set the `rank_refresh_mode` to:
0 (default) = refresh every 5 minutes or so
1 = refresh once on startup only
2 = never refresh

Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` w
 - Minor: Migrated LastFM module to the `reply` response type. (#2118, #2128)
 - Minor: Increased efficiency and speed of subscriber status refresh. (#2203)
 - Minor: Install documentation now recommends the use of limited-scope CloudFlare API tokens. (#2201)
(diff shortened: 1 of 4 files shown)


pajlada:
Previously I've been running my bots with this diff

```
diff --git a/pajbot/managers/user_ranks_refresh.py b/pajbot/managers/user_ranks_refresh.py
index 37108888..03d16eec 100644
--- a/pajbot/managers/user_ranks_refresh.py
+++ b/pajbot/managers/user_ranks_refresh.py
@@ -27,6 +27,7 @@ class UserRanksRefreshManager:
     @staticmethod
     @time_method
     def _refresh(action_queue):
+        return
         try:
             with DBManager.create_dbapi_cursor_scope(autocommit=True) as cursor:
                 cursor.execute("REFRESH MATERIALIZED VIEW CONCURRENTLY user_rank")
```
because the refreshing has a tendency of using too much cpu/storage after refreshing (something something psql bug probably)

with this setting I can run without any local modifications which is great

pajlada:
image

PR #2332, opened 2023-02-17
Remove debugging leftover

DatGuy1:
no changelog entry needed


RAnders00:
That does indeed look like a sensible change :)

pajlada (approved):

PR #2397, opened 2023-03-26
Change minimal supported Python version to 3.9

pajlada:
Pull request checklist:

The big downside to this change is that it's no longer possible to "just" run whatever python comes with your operating system if you're on an older debian/ubuntu LTS system.


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 Remember to bring your dependencies up to date with `./scripts/venvinstall.sh` when updating to this version!
 
+- Breaking: Changed the minimal supported Python version from 3.8 to 3.9. (#2397)
+- Minor: Add native support for the pyenv for managing Python versions. (#2397)
```
Wissididom, line 8: I'm not sure, but this seems to me like it would sound better:
```suggestion
- Minor: Add native support for pyenv for managing Python versions. (#2397)
```

PR #2428, opened 2023-04-09
Migrate to Helix's Badge API

pajlada:
Pull request checklist:


pajbot/apiwrappers/twitch/helix.py
```diff
@@ -991,3 +1046,18 @@ def get_all_chatters(self, broadcaster_id: str, moderator_id: str, authorization
         chatters = list(set(chatters))
 
         return chatters
+
+    def get_channel_badges(self, broadcaster_id: str) -> TwitchBadgeSets:
+        """
+        Calls the Get Channel Chat Badges endpoint https://dev.twitch.tv/docs/api/reference/#get-channel-chat-badges
+        broadcaster_id is a required field, and is the Twitch User ID of the channel whose chat badges you want to get.
```
ALazyMeme, line 1053:
```suggestion
        broadcaster_id is a required field, as is the Twitch User ID of the channel whose chat badges you want to get.
```

pajlada: this feels wrong

RAnders00: I agree with ALazyMeme, another way to word this that both of you might agree with could be "Both [...] and [...] are required parameters."

RAnders00: Yeah, the original is correct, I don't agree with the suggestion. What the comment means is that the broadcaster_id refers to the ID of the channel to get badges for. There's no two separate IDs.

```suggestion
        broadcaster_id is a required field, it specifies the Twitch User ID of the channel whose chat badges you want to get.
```

RAnders00 (approved):

PR #2501, opened 2023-07-23
Fix deprecated usage of redis.hmset

pajlada:
The replacement is to just use redis.hset with the mapping kwarg, as can
be seen in the docs https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.hset

Pull request checklist:

Tested by manually looking at & messing with the `stream_data` redis key and seeing things update as expected


pajlada:
LGTM :+1: 

image

PR #2553, opened 2023-09-24
Fix followage

pajlada:
This requires the bot user to reauthenticate

Pull request checklist:


pajlada:
Works as expected for me

When pulling this, instead of getting no response at all & just an exception in the logs, you now get a good log message + a message similar to `13:28:28 pajbot: @pajlada, Bot does not have permission to check follow age for "pajlada"`

In addition, I tested re-authenticating with the bot and the command started working after that (after a bot restart at least, I didn't check without a bot restart)

pajlada:
image

PR #2624, opened 2024-01-06
Fixed whispers failing breaking commands

pajlada:
This includes a bunch of related-but-unrelated-changes - all the places where we checked the response value in the HTTPErro exception need to explicitly check None
the "not" operator means we would just re-raise it if it's None OR if it has any error, which it does since it raised an exception!

Not stirctly necessary, but now we also print a specific message for whisper 403e

Pull request checklist:


pajlada:
image

PR #2664, opened 2024-02-26
don't filter out disabled commands when rebuilding

DatGuy1:
This inadvertently hid commands that were disabled from the web UI. db_commands already filters out disabled commands, if so desired, through options["enabled"] (currently L353).

(ONLY DONE MINIMAL TESTING)


pajlada (approved):

pajlada:
Thank you!

